PR #16, opened 2019-04-02
Fixing python,java,go guestbook samples to mongoDB<->BE

simonz130:
There was inconsistent bug that BE was trying to talk to Mongo when Mongo was not ready yet due to kubernetes scheduling.

We apply the fix to all guestbook samples (except node that Sean fixed earlier)
https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-initialization/

### DIFF
--- a/golang/go-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
+++ b/golang/go-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
@@ -26,3 +26,7 @@ spec:
           value: "8080"
         - name: GUESTBOOK_DB_ADDR
           value: guestbook-mongodb:27017
+    initContainers:
+        - name: mongo-init
+          image: busybox:1.28
+          command: ['sh', '-c', 'echo $(date) "> checking for envt var for mongo-service"; until nslookup mongo-service; do echo $(date) "> waiting for mongo"; sleep 2; done; echo $(date) "> got envt var for mongo-service"']
--- a/java/java-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
+++ b/java/java-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
@@ -24,3 +24,8 @@ spec:
           value: "8080"
         - name: GUESTBOOK_DB_ADDR
           value: "java-guestbook-mongodb:27017"
+    initContainers:
+        - name: mongo-init
+          image: busybox:1.28
+          command: ['sh', '-c', 'echo $(date) "> checking for envt var for mongo-service"; until nslookup mongo-service; do echo $(date) "> waiting for mongo"; sleep 2; done; echo $(date) "> got envt var for mongo-service"']
+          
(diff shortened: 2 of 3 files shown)


ahmetb:
Your fix won't change the fact that backends are going to talk to mongodb while it's still initializing. So I'm not sure if this fixes whatever issue you're seeing.

This is expected, and we made product fixes to the extension based on this assumption before in the past: This is a common best practice in Kubernetes, it's ok to fail and crash the backend, as it will eventually come back up. 

Lots of users follow this pattern for simplicity. And that's why we have a Ping() method in the initialization of backend: to crash it while mongodb is still not ready.

I don't think doing an initContainer fixes anything. You could also simply fix this with a `readinessProbe` much cleaner running ping command to mongodb. I'm currently -1 for this current solution.

golang/go-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
```diff
@@ -26,3 +26,7 @@ spec:
           value: "8080"
         - name: GUESTBOOK_DB_ADDR
           value: guestbook-mongodb:27017
+    initContainers:
+        - name: mongo-init
+          image: busybox:1.28
+          command: ['sh', '-c', 'echo $(date) "> checking for envt var for mongo-service"; until nslookup mongo-service; do echo $(date) "> waiting for mongo"; sleep 2; done; echo $(date) "> got envt var for mongo-service"']
```
ahmetb: 1. `mongo-service` name sounds incorrect.

2. I've never ever seen in my life a Kubernetes deployment that waits on another service like this.
  The expected behavior is to crash or retry inside the app, not outside in the deployment manifest.

dgageot: I agree with @ahmetb. Plus, this won't work. When the script ends, the mongo pod has an IP but no one can connect to it yet. I've pushed a PR to fix that in nodeJs. The idea, as @ahmetb says, is to retry multiple time the connection to the db.

ahmetb: @dgageot We added handling for crashing a few times into "Cloud Code: Deploy" action. So I'm not sure if retries are necessary. We can rely on Kubernetes restarts here, and a lot of Kubernetes users do too.

When the mongodb.Ping() method causes backend to crash during the initialization, the "Cloud Code: Deploy" demo still works (which is why I'm not sure what's the need for this fix).

simonz130: What if you deploy this sample not using Cloud Code?
I prefer the code to reflect best practices in addition to the extension.

ahmetb: It deploys just fine maybe after restarting 2 times, which is perfectly acceptable in kubernetes. You introduce more failure modes and "look ready" while you're retrying whereas you arent.

simonz130:
It doesn't solve the problem. We won't commit this.

PR #25, opened 2019-04-02
Adding readiness probes for Node.js sample

simonz130:
- Adding readiness probes to node guestbook sample
- Add thresholds for readiness probe

### DIFF
--- a/nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
+++ b/nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
@@ -40,6 +40,8 @@ spec:
           httpGet:
               port: 8080
               path: /messages
-          failureThreshold: 3
-          timeoutSeconds: 2
-          successThreshold: 2
+          periodSeconds: 3
+          initialDelaySeconds: 2
+          failureThreshold: 1
+          timeoutSeconds: 1
+          successThreshold: 1
--- a/nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
+++ b/nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
@@ -36,3 +36,10 @@ spec:
               key: password
         - name: MONGO_HOST
           value: "mongo-service"
+        readinessProbe:
+          httpGet:
+              port: 8080
+              path: /messages
+          failureThreshold: 3
+          timeoutSeconds: 2
+          successThreshold: 2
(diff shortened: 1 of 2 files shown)


nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
```diff
@@ -36,3 +36,10 @@ spec:
               key: password
         - name: MONGO_HOST
           value: "mongo-service"
+        readinessProbe:
+          httpGet:
+              port: 8080
+              path: /messages
+          failureThreshold: 3
```
ahmetb: maybe reduce threshold to 1 so that app crashes right away and Kubecode isn't tricked into thinking "app has stayed up for N seconds" so it sees the crash and keeps waiting until the pod is ready. (if you make it 1, make `initialDelaySeconds: 2`)

that said, if you validated this works OK for the demo and it's successful, keep it.

nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
```diff
@@ -36,3 +36,10 @@ spec:
...
+          timeoutSeconds: 2
+          successThreshold: 2
```
ahmetb: Keep successThreshold:1, if it succeeds once, there's no need to wait another 10s to call the pod ready.

nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
```diff
@@ -36,3 +36,10 @@ spec:
               key: password
         - name: MONGO_HOST
           value: "mongo-service"
+        readinessProbe:
+          httpGet:
```
ahmetb: probably the most important value you can customize is `periodSeconds`. it's 10s default. so if a check fails, it'll take 10s for another check to happen.

this is too slow for a keynote demo. make it like 3s. and make `timeoutSeconds: 1`. if the app can't connect to db within 1s, it's probably gonna magically succeed in the 2nd second.

nodejs/nodejs-guestbook/kubernetes-manifests/guestbook-backend.deployment.yaml
simonz130: It never crashed for me with these settings - neither at docker for desktop nor at gcp

ahmetb: it's because it's set to 3, and mongodb starts within 3x10 seconds.

As long as kubecode didn't call the deployment "Ready" and gave you a URL within that timeframe, it's ok.

PR #30, opened 2019-04-03
Upgrade to latest version of skaffold

dgageot:
Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/golang/go-guestbook/skaffold.yaml
+++ b/golang/go-guestbook/skaffold.yaml
@@ -1,4 +1,4 @@
-apiVersion: skaffold/v1beta6
+apiVersion: skaffold/v1beta7
 kind: Config
 build:
   artifacts:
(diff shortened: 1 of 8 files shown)


simonz130:
1) Will it trigger installing newest skaffold for all the folks that use this in demos, etc?
2) Do we need to do it manually every time there is a new version of skaffold?
3) What happens if we don't do it and folks still update to newest skaffold? Is it all backward compatible?

dgageot:
@simonz130 you're right that it'll require people to upgrade their skaffold if they are not using the latest one. If we don't upgrade, that's ok: skaffold will just print a warning (which we might end up removing).

The reason why I did this is that there's been a huge effort to improve the descriptions on recent schemas. But, now that I double check, `v1beta6` already has those improvements applied.

simonz130:
Thanks David!
Let's postpone this update then for a few weeks.

simonz130:
Closing this as we've updated to v1beta16 here: https://github.com/GoogleCloudPlatform/cloud-code-samples/pull/155

PR #33, opened 2019-04-03
Updating hello world readme for python to point to GCP documentation

simonz130:


python/python-hello-world/README.md
```diff
@@ -7,129 +7,48 @@
...
+
 
 ---
 
 ## Using the Command Line
 As an alternative to using the Cloud Code extension, the application can be deployed to a cluster using standard command line tools
 
-### Using Skaffold
+### Skaffold
 [Skaffold](https://github.com/GoogleContainerTools/skaffold) is a command line tool that can be used to build, push, and deploy your container images as you work
```
ahmetb: this says "as you work" but the cmd below is not "skaffold dev". What's the value of documenting skaffold here?

python/python-hello-world/README.md
```diff
@@ -7,129 +7,48 @@
...
 
 ```
 skaffold run --default-repo=gcr.io/your-project-id-here/cloudcode
 ```
 
-### Using kubectl
+### kubectl
 
 [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) is the official Kubernetes command line tool. It can be used to deploy Kubernetes manifests to your cluster, but images must be build seperately using another tool (for example, using the [Docker CLI](https://docs.docker.com/engine/reference/commandline/cli/))
```
ahmetb: (sorry couldn't comment outside the diff but) 

delete "Build container images" section below, doesn't make any sense given there's already skaffold.

similarly, delete the "Deploy manifests" section too.

python/python-hello-world/README.md
```diff
@@ -7,129 +7,48 @@
...
+
+## Cloud Code for Visual Studio Code
+1. [Getting Started](#getting-started])
+2. [What's in the box](https://cloud.google.com/code/docs/vscode/quickstart#whats_in_the_box)
+3. Using Cloud Code
+    * [Set up a Google Kubernetes Engine Cluster](https://cloud.google.com/code/docs/vscode/quickstart#creating_a_google_kubernetes_engine_cluster)
+    * [Deploy the app](https://cloud.google.com/code/docs/vscode/quickstart#deploying_your_app)
+    * [Continuous Deployment](https://cloud.google.com/code/docs/vscode/quickstart#initiating_continuous_deployment)
+    * [View Container Logs](#view-container-logs)
+    * [Debug Your Code](#debug-your-code)
```
ahmetb: I think you need cloud.google.com links for these items too.

python/python-hello-world/README.md
simonz130: In case folks land on these templates via github and prefer to use the code as a sample without vscode or intellij

python/python-hello-world/README.md
simonz130: SG

python/python-hello-world/README.md
simonz130: Good catch. Thanks!

python/python-hello-world/README.md
ahmetb: still sounds inaccurate. maybe change command to "skaffold dev" below, or edit the descriptio.

IMO folks landing on these templates without using vscode/intellij is a very very small percentage that's not worth optimizing for.

these extra sections are probably making more confusing for the main audience to begin with.

simonz130:
I'm merging and will update instructions for "skaffold delete" in later PRs.

PR #51, opened 2019-04-25
 Adds sensible .gitignore files to java samples

etanshaul:


etanshaul:
@simonz130 fyi

java/java-guestbook/.gitignore
```diff
@@ -0,0 +1,7 @@
+**/.DS_Store
+.idea
```
simonz130: should we include .idea? How will we add changes that trigger extension installation and so on?

java/java-hello-world/.gitignore
```diff
@@ -1 +1,7 @@
-bin/
+**/.DS_Store
+.idea
```
simonz130: ditto here

java/java-guestbook/.gitignore
etanshaul: yeah, you're right. fixed

PR #50, opened 2019-04-25
Change jdwp port to 5005

etanshaul:
Fix incorrect jdwp port

PR #54, opened 2019-04-26
Fix debug port in Java's Dockerfile

quoctruong:
Change the ports in the Dockerfile to match the ones in "launch.json" file.

### DIFF
--- a/java/java-guestbook/.vscode/launch.json
+++ b/java/java-guestbook/.vscode/launch.json
@@ -9,7 +9,7 @@
       "request": "attach",
       "name": "Debug on Kubernetes (frontend)",
       "language": "Java",
-      "debugPort": 50005,
+      "debugPort": 5005,
       "localRoot": "${workspaceFolder}/frontend",
       "remoteRoot": "/app",
       "podSelector": {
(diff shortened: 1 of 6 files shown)


iantalarico:
Why are we switching these? Isn't 5005 the standard port?

@etanshaul @ivanporty

etanshaul:
They should be 5005. This is the standard. I just made this change. Sorry - I should have checked to see how this affects VS Code.

etanshaul:
https://github.com/GoogleCloudPlatform/cloud-code-samples/pull/50

etanshaul:
Fwiw I don’t think it really matters, as long as it’s not in the privileged port range. But 5005 (not 50005) is what I normally see as the default.

iantalarico:
Looks like we are using 50005 in our defaults, we should probably change this given that 5005 is the default.  We may want to revert this and change it back as a fast follow of a Cloud Code VS Code release?

@quoctruong How does this affect our bits if we leave it as 5005?

etanshaul:
> We may want to revert this and change it back as a fast follow of a Cloud Code VS Code release?

This is fine. We can remember to update this after the next VS Code release (I assume this is breaking debugging for the Java apps - sorry again..). FWIW, once we all integrate `skaffold debug` we can get rid of this :)

quoctruong:
@iantalarico @etanshaul PTAL. I changed all the ports to 5005 instead.

etanshaul:
Oh right! the vs code launch configs live here in the samples, not the extension. So we don't have to wait for a release.

quoctruong:
Thanks!

PR #59, opened 2019-05-31
Add Django Hello World template

wangxf123456:
The Django template includes the following files:
* .idea:
    * Copied from python/python-hello-world/.idea
* .vscode
    * Copied from python/python-hello-world/.vscode
    * Modify tasks.json to make it use Python3 to run pylint because the Django template is generated by [Django 2.2.1](https://docs.djangoproject.com/en/2.2/releases/2.0/#python-compatibility), which only supports Python 3
    * Modify launch.json to make it work for the Django template
* kubernetes-manifests
    * Includes a Django server container, a Postgres container and a Job to run database migration. The Django application is incomplete without database migration, even for a hello world app, because all Django application comes with its [admin interface](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#module-django.contrib.admin), which uses the database
* src
    * Generated by "django-admin [startproject](https://docs.djangoproject.com/en/2.2/ref/django-admin/#startproject) helloworld" + Dockerfile + requirements.txt + Modification of Django files to use the Postgres database
* skaffold.yaml

I tested the template on my Linux workstation with a GKE cluster with the following commands:
* vscode:
    * Cloud Code: Deploy
    * Cloud Code: Continuous Deploy
* Intellij:
    * Kubernetes Continuous Deploy
    * Kubernetes Deploy

I found the way to do database migration with Kubernetes on this [tutorial](https://www.agiliq.com/blog/2018/07/django-on-kubernetes/).


wangxf123456:
@simonz130 @brianquinlan @enriquejosepadilla Looks like I am not able to add reviewers to this PR.

.templates/template_location.json
```diff
@@ -11,6 +11,12 @@
     "templatePath": "python-guestbook",
     "templateName": "Python: Guestbook"
   },
+  {
+    "repoPath": "https://github.com/GoogleCloudPlatform/cloud-code-samples.git",
```
iantalarico: Adding this will add this immediately to the the Cloud Code extensions once checked it.  Are you okay to hold off on this for a bit so we can do a little more testing?

wangxf123456: Sure. Thanks for the reminder.

python/django/python-hello-world/src/requirements.txt
```diff
@@ -0,0 +1,2 @@
+Django==2.2.1
```
simonz130: please add a brief readme file that focuses on:
1) How to build
2) How to debug
3) How to deploy
All could be links to existing cloud.google.com documentation.
It should also show the sample app "architecture" (maybe worth highlighting django involvement in the sample)

You can look at existing hello-world python sample readme for inspiration/

wangxf123456: I added a readme file, and this readme file is basically the same with the readme file in the existing hello-world python sample, except the diagram showing the structure of the template, and the description of the content of the template.

wangxf123456:
I modified the Django files so that it prints "Hello World" on the main page. Previously it shows a web page which looks like the following, but this is inconsistent with other templates.  
 
00_welcome

python/django/python-hello-world/README.md
```diff
@@ -0,0 +1,72 @@
...
+    * .vscode
+      * extensions.json: prompt requesting download of related extensions when opening this project
+      * launch.json: information needed for the debugger to attach to our service
+      * tasks.json: configuration information for Visual Studio Code Tasks
+    * kubernetes-manifests
+      * hello.deployment.yaml: Kubernetes manifest for the Hello World server Deployment
+      * hello.service.yaml: Kubernetes manifest for the Hello World LoadBalancer Service
+      * hello.migration.yaml: Kubernetes manifest for the Postgres database migration job
+      * postgres.development.yaml: Kubernetes manifest for the Postgres database Deployment
+      * postgres.service.yaml: Kubernetes manifest for the Postgres database Cluster IP Service
```
simonz130: why do we need these if it's just "hello-world"?

python/django/python-hello-world/README.md
```diff
@@ -0,0 +1,72 @@
+# Hello World with Cloud Code
+
+![Architecture Diagram](./img/diagram.png)
```
simonz130: This diagram is great for hello-world use-case but you also have database. I suspect that you mixed two templates into one:
 - hello-world sample should only print "hello world" and have no dependencies.
 - guestbook sample can depend on postgres or other storage and allow storing and retrieving guestbook records

python/django/python-hello-world/kubernetes-manifests/hello.deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+        app: python-hello-world
+        tier: app
+    spec:
+      initContainers:
+      # Wait for the database to get ready before starting the server, otherwise
+      # the server will crash
+      - name: check-db-ready
+        image: postgres
+        command: ['sh', '-c',
+          'until pg_isready -h python-hello-world-db.default.svc.cluster.local -p 5432;
```
simonz130: why do you need this for hello-world?

python/django/python-hello-world/kubernetes-manifests/hello.migration.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  template:
+    metadata:
+      name: python-hello-world-migration
+    spec:
+      initContainers:
+      # Wait for the database to get ready before doing migration
+      - name: check-db-ready
+        image: postgres
+        command: ['sh', '-c',
```
simonz130: ditto this

python/django/python-hello-world/src/helloworld/settings.py
```diff
@@ -0,0 +1,123 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/2.2/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '#&lxs78@rm-p!-)0g8adrf*=&*ngvaqdp^j(tv#h9p%a!_yxna'
```
simonz130: what is this?

python/django/python-hello-world/src/helloworld/settings.py
```diff
@@ -0,0 +1,123 @@
...
+]
+
+WSGI_APPLICATION = 'helloworld.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/2.2/ref/settings/#databases
+
+
+DATABASES = {
```
simonz130: why is this needed for hello-world?

python/django/python-hello-world/README.md
wangxf123456: Yeah I agree that it sounds strange for a hello world application to require a database. But looks like Django was designed to produce database-backed web sites.

Django has a command to create a project template "[startproject](https://docs.djangoproject.com/en/2.2/ref/django-admin/#startproject)". In the project created with this command, by default it uses a sqlite database, though the project has nothing. The reason is that Django projects have a built in [admin interface](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#module-django.contrib.admin) by default, which uses database.

I can remove the admin interface from the template so that databases are not required, but this may look strange for Django developers. Why the admin interface is enabled in the default project template used by startproject, but not in our template?

python/django/python-hello-world/src/helloworld/settings.py
wangxf123456: The SECRET_KEY is used to provide Cryptographic signing, for example, password reset tokens, sessions. See https://docs.djangoproject.com/en/2.2/ref/settings/#secret-key for more details.

Django refuses to start if SECRET_KEY is not set, so I must include a SECRET_KEY here. But the SECRET_KEY should be unique for each Django application, so I modified the value of the SECRET_KEY to indicate users to modify its value. I also mentioned this in README.

python/django/python-hello-world/src/helloworld/settings.py
wangxf123456: The Django application needs a database because it has an admin interface. By default the settings include a sqlite database. 
```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
    }
}
```

Using sqlite for deployment is inappropriate, so I changed it with postgres.

python/django/python-hello-world/README.md
simonz130: I think that if we can remove the admin interface for hello world, we should do it.
It should be as simple as possible template. For guestbook we should leave admin interface as this will show second level of complexity. Also this will be great opportunity for external developers to see us show both options

python/django/python-hello-world/README.md
```diff
@@ -0,0 +1,75 @@
...
+    * [kubectl](#using-kubectl)
+
+----
+
+### Getting Started
+
+This sample was written to demonstrate how to use the Cloud Code extension for Visual Studio code.
+
+Please change [SECRET_KEY](./src/helloworld/settings.py#L28) as soon as possible. If you keep
+its value unchanged, your application will be vulnerable from security issues. See [link](https://docs.djangoproject.com/en/2.2/ref/settings/#secret-key) for more details.
```
simonz130: Can you be more concrete in the message? What is the security thing that developers risk being exposed if they don't change SECRET_KEY?

simonz130: Also, can we get rid of it altogether? All of the other samples don't require changing anything to get started.

python/django/python-hello-world/src/helloworld/settings.py
```diff
@@ -0,0 +1,129 @@
...
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'postgres',
+        'USER': 'postgres',
+        'HOST': 'python-hello-world-db',
+        'PORT': '5432',
```
simonz130: If we end up having database in the sample, which I really hope we don't, you should read the port from environment variable that is set by kubernetes ($PORT) instead of hardcoding. Otherwise it's possible that folks will change yaml config to use a different port and won't notice that it requires a change here too and the sample might stop working.

python/django/python-hello-world/README.md
wangxf123456: Sure I will the admin interface.

python/django/python-hello-world/README.md
wangxf123456: The security issues are described in the link I provided. Should I also describe them here? 

Actually even if the users do not modify SECRET_KEY, the application will still run successfully. Here I mean if they would like to start development based on the template, they should modify the SECRET_KEY as soon as possible. I will modify the message to make this more clear.

python/django/python-hello-world/kubernetes-manifests/hello.deployment.yaml
wangxf123456: Removed database for hello-world template.

python/django/python-hello-world/kubernetes-manifests/hello.migration.yaml
wangxf123456: Removed database for hello-world template.

python/django/python-hello-world/src/helloworld/settings.py
wangxf123456: Removed the database settings.

PR #62, opened 2019-06-11
Add Django Guestbook template

wangxf123456:
This one is similar with the Hello World template, but a Postgres database is included.


wangxf123456:
@simonz130 @brianquinlan @enriquejosepadilla

wangxf123456:
@simonz130 Friendly ping.

python/django/python-guestbook/kubernetes-manifests/guestbook-app.deployment.yaml
```diff
@@ -0,0 +1,41 @@
...
+        app: python-guestbook
+        tier: app
+    spec:
+      initContainers:
+      # Wait for the database to get ready before starting the server, otherwise
+      # the server will crash
+      - name: check-db-ready
+        image: postgres
+        command: ['sh', '-c',
+          'until pg_isready -h python-guestbook-db.default.svc.cluster.local -p 5432;
```
simonz130: can you replace the port with environment variable holding the port number? I see you have the port in DATABASE_PORT

python/django/python-guestbook/kubernetes-manifests/guestbook-app.migration.yaml
```diff
@@ -0,0 +1,34 @@
...
+  template:
+    metadata:
+      name: python-guestbook-migration
+    spec:
+      initContainers:
+      # Wait for the database to get ready before doing migration
+      - name: check-db-ready
+        image: postgres
+        command: ['sh', '-c',
+          'until pg_isready -h python-guestbook-db.default.svc.cluster.local -p 5432;
```
simonz130: ditto

simonz130:
> @simonz130 Friendly ping.

Sorry for the delay, I was out for 2 weeks and now back.
LGTM from me but I also asked Charlie, our Python guru to take a second look.

python/django/python-guestbook/kubernetes-manifests/guestbook-app.deployment.yaml
wangxf123456: Thanks very much for your feedback. I tried to replace "5432" with "$DATABASE_PORT", but looks like "containers" and "initContainers" are in different environments. Those environment variables are only available in "containers" but not in "initContainers". I tried to find whether there are any ways to define environment variables for both "containers" and "initContainers" but was not able to find any.

simonz130: OK,  let's leave it as 5432 then. Thanks for checking the options.

PR #72, opened 2019-08-02
Add IntelliJ metadata to the Python projects to fix initial loading issue

etanshaul:
## What this PR does

Adds the intellij metadata (in the `.idea` folder) to version control.

## Why?

Given the current state of the python samples, when opened in IntelliJ for the first time, the following appears:
image

The project isn't correctly detected, and the user remains in this odd state.

Closing and re-opening the project fixes things, then the user sees the correct structure (and can use Cloud Code):
image

Obviously this is not ideal.

By committing the metadata, and having it available to the IDE on first load, this problem is fixed, and the project loads properly on the first try and can immediately use Cloud Code features.

Note: Tested working in Idea Ultimate (with the python plugin), and in Pycharm


etanshaul:
note: a similar fix will be needed for `Go` and possibly (not-yet-confirmed) for `Node.js`. Will go one by one per language.

etanshaul:
@patflynn as fyi since I think you may have been running into something similar

.templates/template_location.json
```diff
@@ -3,14 +3,26 @@
     "repoPath": "https://github.com/GoogleCloudPlatform/cloud-code-samples.git",
     "directoryPath": "python/python-hello-world",
     "templatePath": "python-hello-world",
-    "templateName": "Python: Hello World"
+    "templateName": "Python (Flask): Hello World"
```
simonz130: Could you please rebase your local branch so template_location.json is not part of your change?
I wanted to submit this change separately (addition of django)

python/django/python-hello-world/.idea/misc.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectRootManager" version="2" project-jdk-name="Python 3.6" project-jdk-type="Python SDK" />
```
simonz130: what is this?
Is it native intellij setting? Should we maintain this when/if we change the version of python supported in the templates?

.templates/template_location.json
etanshaul: Oops. Didnt catch that. Will update.

python/django/python-hello-world/.idea/misc.xml
etanshaul: Good question. Let me test without it to see if it still works

etanshaul:
@simonz130 ptal

Merged from upstream, and removed the `misc.xml` files.

etanshaul:
@googlebot I fixed it..

etanshaul:
@simonz130 could you please merge for me once approved? I don't have permissions

python/django/python-hello-world/.idea/misc.xml
etanshaul: fixed. files removed

.templates/template_location.json
etanshaul: fixed. merged from upstream

PR #73, opened 2019-08-05
Adding Jib based Java building.

patflynn:
Uses [JIB](https://github.com/GoogleContainerTools/jib) as the default skaffold builder for Java

Hi there! I'm the TL for Cloud Code and we'd prefer to use Jib as the Java builder. Jib builds without the need for local Docker or a project Dockerfile, and considerably improves build speed.  

I've included a comment for those who want to use the included Dockerfile.


patflynn:
@loosebazooka FYI

patflynn:
FYI I'm going to update this to use skaffold profiles to pick build strategy

java/java-hello-world/pom.xml
```diff
@@ -46,6 +46,12 @@
 	</dependencies>
 	<build>
 		<plugins>
+
```
simonz130: nit - remove empty line

java/java-hello-world/skaffold.yaml
```diff
@@ -5,8 +5,13 @@ build:
     sha256: {}
   # defines where to find the code at build time and where to push the resulting image
   artifacts:
-  - image: java-hello-world
-    context: .
+    - image: java-hello-world
+      context: .
+      # Comment out the jibMaven line below to use the Dockerfile based build.
+      # To learn more about how jib build Java containers visit
```
simonz130: nit: jib builds?

java/java-hello-world/skaffold.yaml
```diff
@@ -5,8 +5,13 @@ build:
...
+      # https://github.com/GoogleContainerTools/jib
+      jibMaven: {}
```
simonz130: Will Cloud Code download Jib on the machines where it's not installed? Will it work both on Windows, macOS and Linux the same?

loosebazooka: yes, jib is just a maven/gradle plugin, those systems handle downloading everything required.

loosebazooka:
Should we include a mvn wrapper for the maven dependency?

etanshaul:
> Should we include a mvn wrapper for the maven dependency?

I definitely think so. But sounds like a separate PR?

java/java-hello-world/skaffold.yaml
patflynn: fixed

patflynn:
updated guestbook app and added dockerfile profiles

patflynn:
@simonz130 I just looked at the cloud code docs and they don't seem to go into how the containers are built. Should be good to make this change without any update there.

patflynn:
thanks @simonz130 !

PR #75, opened 2019-08-05
Add IntelliJ metadata for node projects

etanshaul:
See #74 and #72 

This is the same idea, but for the NodeJS projects


etanshaul:
@simonz130 ptal

nodejs/nodejs-hello-world/.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/nodejs-hello-world.iml" filepath="$PROJECT_DIR$/.idea/nodejs-hello-world.iml" />
```
simonz130: Are both parameters needed?

etanshaul: You mean “fileurl” and “filepath”? I’m not totally sure, but these files were auto generated by the IDE so I lean towards leaving them as is.

simonz130: Yeah, if it's auto-generated, let's leave them. I'm merging your PR.

Thanks!

PR #78, opened 2019-08-06
Add template language metadata to template_location.json

etanshaul:
Adding a `templateLanguage` array to each template metadata entry in `template_location.json`.

## Why?

The IDEs may require this knowledge in order to determine which templates to make available to the user in certain contexts. For example, if the user is running in GoLand, we only want to show the `go` templates to the user, as the other ones won't have any language support provided by the IDE.

## Why an array?

Future templates could be multi-service applications with multiple languages, such as [Hipster Shop](https://github.com/GoogleCloudPlatform/microservices-demo)

### DIFF
--- a/.templates/template_location.json
+++ b/.templates/template_location.json
@@ -3,60 +3,70 @@
     "repoPath": "https://github.com/GoogleCloudPlatform/cloud-code-samples.git",
     "directoryPath": "python/python-hello-world",
     "templatePath": "python-hello-world",
-    "templateName": "Python: Hello World"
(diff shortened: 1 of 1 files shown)


etanshaul:
@redborian for FYI / comment from the VS Code side.

etanshaul:
please hold on merging until confirmation from VS Code that this won't break their parsing

etanshaul:
@simonz130 synced with upstream. Should be good to merge now. Thanks.

etanshaul:
friendly ping. could we get this merged?

PR #81, opened 2019-08-14
Add templateLanguage to the .template README

etanshaul:


.templates/README.md
```diff
@@ -3,3 +3,4 @@
 * `directoryPath`: Since a repository can contain multiple templates, `directoryPath` is used to determine the location of the template in a repository.
 * `templatePath`: Name of the folder where the template will be downloaded too.
 * `templateName`: Name of the template.
+* `templateLanguages`: The list of programming languages used by the template.
```
simonz130: Please specify which languages (i.e. values for this field) are currently supported.

etanshaul: are you sure we want this? It creates another location where we have to remember to update this as we add new templates with new languages. Similar to how we don't enumerate all of the `templateName`'s that we have.

simonz130: Template name is free form. The reason I'm thinking to have pre-defined values for languages is that we consider improving the **Cloud Code: New App** experience to first ask for a language, and then show the templates for that language only. Now it's not a problem but once more samples are added, it might become infeasible to show all of the templates in the same dropdown, at which point pre-set values will help to filter by language.

etanshaul: fair enough. Although having them here in the README does nothing to enforce that the values in `templateLanguage` are not freeform.

In any case, I don't feel to strongly about it. wdyt about this:

```
`templateLanguages`: The list of programming languages used by the template.  Current possible values include:
  * `java`
  * `python`
   ....
```

simonz130: Yeah, this list example looks good.
Agree that we need it to be enforced and not just in readme. We'll do it step-by-step.

etanshaul: cool. updated

PR #92, opened 2019-09-18
Fixing debian caching issue for .net guestbook FE

simonz130:


simonz130:
Currently disabling CI for dotnet guestbook until solution for handling dependencies for debugging is found.

PR #91, opened 2019-09-18
Add IDEA metadata to Java sample projects

etanshaul:
Same concept as #75 #74 #72 but for Java samples

PR #96, opened 2019-09-24
Fix for java-hello-world iml IDEA metadata

etanshaul:
Follow up fix for #91 

The path in `modules.xml` is pointing to an `iml` file under `.idea` called `java-hello-world.iml`. This PR moves the previously incorrectly named (and wrongly located) iml file to the expected location so that the IDE doesn't complain when loading the project. We were seeing these messages:

```
.../java-hello-world/.idea/java-hello-world.iml does not exist Please correct the file content when I try to use the wizard flow.
```

similar fix for the Guestbook Java app coming as well.


etanshaul:
@simonz130 ptal

PR #97, opened 2019-09-24
Fix for java-guestbook sample iml metadata location

etanshaul:
Same fix as #96 but for the guestbook app


etanshaul:
@simonz130 ptal. thanks!

etanshaul:
@simonz130 would be great if we could have this merged today for testing (ahead of the Cloud Code IntelliJ release next Monday). Should be identical to the already merged #96.

PR #102, opened 2019-10-02
Simplify Golang images

briandealwis:
- Add `GOTRACEBACK` env var so `skaffold debug` can identify images as golang
- Use multi-stage builds to produce smaller images (as [per distroless[(https://github.com/GoogleContainerTools/distroless#examples-with-docker))


briandealwis:
We need some way to identify the language runtime.  There’s an open issue about defining an explicit mechanism (https://github.com/GoogleContainerTools/skaffold/issues/2203),  but this implicit environment approach is effective.

PR #153, opened 2019-10-24
Fix #152 for multimodule projects

### LINKED ISSUE #152: Java Multimodule Guestbook App is incorrectly configured for Jib
In a Jib multimodule project, the context directory should point to the root and the project should be the subproject location.

briandealwis:
- top-level project should have `<packaging>pom</packaging>` since it doesn't produce a jar
- fix specification of Skaffold/Jib multimodule projects: the artifact `context` should point to the root of the maven project (`.` in this case, and so `context` is now omitted), and the `module` should be the relative directory of the sub-project/module

### DIFF
--- a/java/java-guestbook/pom.xml
+++ b/java/java-guestbook/pom.xml
@@ -7,6 +7,7 @@
     <groupId>com.cloudcode.guestbook</groupId>
     <artifactId>java-guestbook</artifactId>
     <version>1.0</version>
+    <packaging>pom</packaging>
 
     <modules>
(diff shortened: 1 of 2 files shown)


ivanporty:
@briandealwis I just opened the larger PR (#154) where everything addressed all together, and schema is updated as well to make sure people are using the latest jib configuration block when they start with samples. PTAL and if you think all is good there, let's close this one then?

briandealwis:
Closing in favour of #154

Issue #157, opened 2019-10-25
Samples are broken for Cloud Code IntelliJ after update to vbeta16


ivanporty:
CC IJ provides managed dependencies, and the latest schema supported by managed Skaffold is vbeta15, not 16. We are planning release for Oct 31st tentatively, need to revert the schema update back to `v1beta15` until it's in.

simonz130:
I'll downgrade the schema to 15.

Is there a way we can test IntelliJ samples flow automatically?

ivanporty:
Not yet unfortunately. Target Q4-Q1.

ivanporty:
(for initial set of integration testing.)

ivanporty:
@sageprice FYI, is there any option to test the samples against our latest managed dependencies versions of skaffold and kubectl? Maybe download and test samples with them?

sageprice:
> @sageprice FYI, is there any option to test the samples against our latest managed dependencies versions of skaffold and kubectl? Maybe download and test samples with them?

AFAIK this would have to be done manually as you've described. If we want to test the samples against skaffold/kubectl versions newer than what we're using then we have to download the latest and specify their usage as custom dependencies in settings:
![custom-deps](https://user-images.githubusercontent.com/9577120/67589865-86458900-f727-11e9-8f54-695fcfe2e47c.png)

simonz130:
How IntelliJ tests "Create new application" flow today? Is it done manually for each release?

ivanporty:
Yes, it is manual. Each new release has a fixed version of `skaffold` bundled and used by Cloud Code managed dependencies component. It will be same for VS Code once they release managed dependencies. We test manually mostly now, but again, planning to get some integration coverage soon.

simonz130:
Thanks Ivan. I'm closing the bug for now as the samples should not be broken at this point. We should discuss it further how to automate the testing.

PR #158, opened 2019-10-25
Update skaffold schema to v1beta15

simonz130:
This will mitigate the issue visible in intellij when working with the the templates


simonz130:
Solves #157

ivanporty:
@simonz130 this should be good to go, but CI fails - I think some problem with Python samples after the latest upgrades?

ivanporty:
But anyway, regardless of the CI results, we need to get this in ASAP, currently essentially quick start / samples flow is completely broken for CC IJ.

simonz130:
Merging to unblock IntelliJ and will open a separate issue on CI failure.

PR #172, opened 2019-11-07
remote unnecessary files in hello-world samples

simonz130:


simonz130:
Merging directly as these files are not involved in CI and we currently work on resolving CI issues.

PR #177, opened 2019-11-09
Fix python django app

quoctruong:
Python Django App is not configured for debugging so we have to add support for it.

### DIFF
--- a/python/django/python-guestbook/.vscode/settings.json
+++ b/python/django/python-guestbook/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "cloudcode.profile-cluster-map": {
+        "[default]": "another-context"
+    },
+    "cloudcode.profile-registry-map": {
+        "[default]": "gcr.io/gcloud-powershell-testing"
+    }
+}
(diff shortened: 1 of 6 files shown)


python/django/python-guestbook/.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "cloudcode.profile-cluster-map": {
+        "[default]": "another-context"
```
simonz130: can we use more meaningful name?

python/django/python-hello-world/.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "cloudcode.profile-cluster-map": {
+        "[default]": "another-context"
```
simonz130: Please remove the settings file

python/django/python-guestbook/.vscode/settings.json
simonz130: or actually, remove the settings file changes altogether

quoctruong:
@simonz130 PTAL

simonz130 (approved):

PR #201, opened 2020-01-10
Make samples objects consistent across languages

simonz130:
- Changed go and node samples to comply with the same pattern for object naming.
Example: *go-hello-world* instead of hello-world.
This will help to use the same template for CI across all the languages without IF statements that check for special cases (like it is today)

### DIFF
--- a/golang/go-guestbook/kubernetes-manifests/guestbook-mongodb.deployment.yaml
+++ b/golang/go-guestbook/kubernetes-manifests/guestbook-mongodb.deployment.yaml
@@ -6,7 +6,7 @@
 kind: Deployment
 apiVersion: apps/v1
 metadata:
-  name: go=guestbook-mongodb
+  name: go-guestbook-mongodb
   labels:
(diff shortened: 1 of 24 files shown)


golang/go-guestbook/kubernetes-manifests/guestbook-mongodb.deployment.yaml
```diff
@@ -6,7 +6,7 @@
 kind: Deployment
 apiVersion: apps/v1
 metadata:
-  name: guestbook-mongodb
+  name: go=guestbook-mongodb
```
quoctruong: I think this should be `go-` instead of `go=`?

simonz130: Thanks, for the catch!

simonz130:
Finally, all the builds are passing with the new triggers setup.

@quoctruong, could you please sign off?

PR #217, opened 2020-02-14
Add Cloud Run (Golang) Hello World

grayside:
This is an initial port of cloud-run-hello for use as a sample.

I've annotated it with a few TODOs/considerations for us to work out further definition of what these should look like.

All SVGs were copied verbatim from the cloud-run-hello sample.

* [ ] Golang Code Review
* [ ] Finalize sample design
* [ ] Complete Cloud Code integration
* [ ] Complete README
* [ ] Switch integration tests to Go instead of BASH


cloudbuild.meta.yaml
```diff
@@ -46,6 +46,7 @@ steps:
     declare -a apps=("hello-world" "guestbook")
     gcloud builds submit . --config cloudbuild.tmpl.yaml --substitutions _DIR=${dir},_LANG=${lang},_APP=${apps[0]}
     gcloud builds submit . --config cloudbuild.tmpl.yaml --substitutions _DIR=${dir},_LANG=${lang},_APP=${apps[1]}
+    gcloud builds submit . --config golang/go-cloud-run-hello-world
```
grayside: Per discussion with Simon, we do not need to use the substitutions since we're using a unique cloudbuild.yaml per sample.

golang/go-cloud-run-hello-world/.gitignore
```diff
@@ -0,0 +1,14 @@
+# Binaries for programs and plugins
```
grayside: This file was copied from the `go-hello-world` sample.

golang/go-cloud-run-hello-world/.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
+/workspace.xml
```
grayside: This file was copied from the go-hello-world sample.

golang/go-cloud-run-hello-world/.idea/externalDependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+    <component name="ExternalDependencies">
+          <plugin id="com.google.gct.core" />
+    </component>
+</project>
```
grayside: This file was copied from the `go-hello-world` sample.

golang/go-cloud-run-hello-world/.idea/go-cloud-run-hello-world.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="WEB_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
grayside: This file was copied from the `go-hello-world` sample.

golang/go-cloud-run-hello-world/.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/go-cloud-run-hello-world.iml" filepath="$PROJECT_DIR$/.idea/go-cloud-run-hello-world.iml" />
+    </modules>
+  </component>
+</project>
```
grayside: This file was copied from the `go-hello-world` sample and the module reference updated to the dirname of this sample.

golang/go-cloud-run-hello-world/.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    {
+      "type": "go",
+      "request": "launch",
+      "name": "Launch (local)",
+      "mode": "auto",
+      "program": "${workspaceFolder}/."
+    },
+  ]
+}
```
grayside: I'm not sure what to do with this file, especially given the discussion to leave skaffold out of the Cloud Run samples.

golang/go-cloud-run-hello-world/.vscode/tasks.json
```diff
@@ -0,0 +1,42 @@
...
+            "command": "kubectl",
+            "args": [
+                "logs",
+                "--selector",
+                "app=hello-world"
+            ],
+            "problemMatcher": [],
+        }
+    ]
+}
```
grayside: I'm not sure what to do with this file, especially given the discussion to leave skaffold out of the Cloud Run samples.

golang/go-cloud-run-hello-world/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# container and point it to the "dlv debug" command:
+#   * UNCOMMENT the following ENTRYPOINT statement,
+#   * COMMENT OUT the last ENTRYPOINT statement
+# Start the "dlv debug" server on port 3000 of the container.
+ENTRYPOINT ["dlv", "exec", "/app", "--continue", "--accept-multiclient", "--api-version=2", "--headless", "--listen=:3000", "--log"]
+
+# If you want to run WITHOUT the debugging server:
+#   * COMMENT OUT the previous ENTRYPOINT statements,
+#   * UNCOMMENT the following ENTRYPOINT statement.
+# ENTRYPOINT ["/app"]
```
grayside: This file was copied from `go-hello-world`.

Changes include:
* adding static assets
* swapping in the -mod=readonly directive used in the Cloud Run samples.

Normally for golang I'd use a multi-stage Dockerfile, but not sure what the right way is to split the difference with local debugging, so I'm following the existing sample. If we're using this file as a model for a production Dockerfile, it might be worth doing something more complicated.

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
...
+  --allow-unauthenticated \
+  --platform managed \
+  --region us-central1 \
+```
+
+## Run the Tests
+
+```
+gcloud builds submit . --config cloudbuild.yaml --substitutions COMMIT_SHA=manual
+```
```
grayside: Started sketching out this README, but it seems like there may be some flux on what we want to include.

golang/go-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,67 @@
...
+  entrypoint: '/bin/bash'
+  args:
+  - '-c'
+  - |
+    get_url() {
+        gcloud run services describe ${_SERVICE} --format 'value(status.url)' \
+          --platform managed --region ${_CLOUDSDK_REGION}
+    }
+    echo $(get_url) > _service_url
+    echo "Cloud Run URL for ${_SERVICE} is $(cat _service_url)"
```
grayside: This was adapted from the `go-hello-world` sample. Since the `test_content.sh` is no longer trying to be general-purpose this could be moved into it to simplify the `cloudbuild.yaml`.

golang/go-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+- id: 'Deploy to Cloud Run'
+  name: 'gcr.io/cloud-builders/gcloud:$_CLOUDSDK_VERSION'
+  args:
+  - 'run'
+  - 'deploy'
+  - '${_SERVICE}'
+  - '--image'
+  - 'gcr.io/${PROJECT_ID}/${_SERVICE_TAG}:${COMMIT_SHA}'
+  - '--allow-unauthenticated'
```
grayside: We should probably remove `--allow-unauthenticated` and instead run the integration tests with an identity token via `gcloud auth print-identity-token`. Kept this here for initial simplicity.

golang/go-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+# Uncomment if skipping teardown to associate build with container image.
+# images:
+# - 'gcr.io/${PROJECT_ID}/${_SERVICE}:${COMMIT_SHA}'
+
+substitutions:
+  _CLOUDSDK_VERSION: latest
+  _CLOUDSDK_REGION: us-central1
+  _SERVICE: hello-world
+  _SERVICE_TAG: go-cloud-run-hello-world
```
grayside: This sample name is really long so I didn't want to use it as the service URL itself if developers might manually type it.

golang/go-cloud-run-hello-world/index.html
```diff
@@ -0,0 +1,207 @@
+<!doctype html>
```
grayside: This file was copied as-is from cloud-run-hello. There are open issues in the repo to tune it for accessibility so we'll want to track and port changes

golang/go-cloud-run-hello-world/main.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 Google LLC
```
grayside: I updated this sample from the original to better align with Cloud Run sample practices, including more error handling and use of the metadata client library.

golang/go-cloud-run-hello-world/test_content.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "START CONTENT TEST"
+echo "retries: "$retries
+echo "interval: "$interval
+echo "keyword: "$keyword
+echo "url: "$url
+
+if [[ -z "$url" ]]; then
+  echo "Failed: URL not specified."
+  exit 1
+fi
```
grayside: I added this stanza because during development of the `cloudbuild.yaml` I had an error where the URL was an empty value and the test process went through all retries for no reason.

golang/go-cloud-run-hello-world/index.html
grayside: Also, this page has gotten complicated as it emphasizes a strong out-of-box experience as a running app, rather than being a sample HTML page. Do we want to maintain this as-is in multiple languages or simplify?

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
...
+This "Hello World" is a [Cloud Run](https://cloud.google.com/run/docs) service that renders a webpage.
+
+----
+## Table of Contents
+
+### Cloud Code for Visual Studio Code
+
+1. [Getting Started](#getting-started])
+2. [What's in the box](https://cloud.google.com/code/docs/vscode/quickstart#whats_in_the_box)
+3. [Using the Command Line](#using-the-command-line)
```
simonz130: We shouldn't have the section of using command line (even though it should work when invoking appropriate command line). This is because we optimize for Cloud Code IDE experience (IntelliJ and vscode).

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
+# Cloud Run Hello World with Cloud Code
+
+This "Hello World" is a [Cloud Run](https://cloud.google.com/run/docs) service that renders a webpage.
+
+----
+## Table of Contents
+
+### Cloud Code for Visual Studio Code
```
simonz130: Our instructions should equally target VSCode and IntellIJ. Let's just remove this header and have "Table of contents" only.

golang/go-cloud-run-hello-world/README.md
simonz130: We probably need two more sections:

1) Running locally
2) Running on Cloud Run

simonz130: Also, maybe "Next steps" with more useful links.

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
...
+* [Install Cloud Code for VS Code](https://cloud.google.com/code/docs/vscode/install)
+* [Creating a new app](https://cloud.google.com/code/docs/vscode/creating-an-application)
+* [Editing YAML files](https://cloud.google.com/code/docs/vscode/yaml-editing)
+
+## Using the Command Line
+
+## Build the Container Image
+
+```
+docker build -t go-cloud-run-hello-world .
```
simonz130: This should be transparent for users of Cloud Code: it will build container image as part of trying to deploy the app locally or to Cloud Run.

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Run the Container Locally
+
+```
+docker run ...
```
simonz130: Likely will turn into:

* VSCode: run **Cloud Code: Run Locally** command on command pallete.
* IntelliJ: something along similar lines

golang/go-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Deploy to Cloud Run
```
simonz130: The actual command will be abstracted away by Cloud Code extension.

golang/go-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - 'managed'
+
+- id: 'Get Cloud Run URL'
+  name: 'gcr.io/cloud-builders/gcloud:$_CLOUDSDK_VERSION'
+  entrypoint: '/bin/bash'
+  args:
+  - '-c'
+  - |
+    get_url() {
+        gcloud run services describe ${_SERVICE} --format 'value(status.url)' \
```
simonz130: We need to have isolation for these templates: in case multiple CI runs are happening at the same time, they shouldn't have conflicts w.r.t names. In kubernetes it's easy to achieve with namespaces. Not sure how to do it with Cloud Run.

golang/go-cloud-run-hello-world/index.html
simonz130: When you say simplify - do you mean having the same template HTML referenced in multiple languages templates?

golang/go-cloud-run-hello-world/main.go
```diff
@@ -0,0 +1,101 @@
...
+
+package main
+
+import (
+	"html/template"
+	"log"
+	"net/http"
+	"os"
+
+	"cloud.google.com/go/compute/metadata"
```
simonz130: How this will work when the template will be running locally?

golang/go-cloud-run-hello-world/main.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Variables used to generate the HTML page.
+var (
+	data templateData
+	tmpl *template.Template
+)
+
+func main() {
+	// Comment out in local development for timestamp prefixes.
```
simonz130: Can this be implemented such that there will be no required manual step to get this to run locally vs on Cloud Run?

golang/go-cloud-run-hello-world/main.go
```diff
@@ -0,0 +1,101 @@
...
+	// Define HTTP server.
+	http.HandleFunc("/", helloRunHandler)
+
+	fs := http.FileServer(http.Dir("./assets"))
+	http.Handle("/assets/", http.StripPrefix("/assets/", fs))
+
+	// PORT environment variable is provided by Cloud Run.
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
```
simonz130: Do we need to mention in the readme that it runs on 8080 when deployed locally?

golang/go-cloud-run-hello-world/index.html
grayside: The current webpage involves multiple images, complicated CSS, support for dark mode, etc. It is complicated enough that it's had a number of changes recently to improve accessibility, performance, and so on: in other words, it's not really maintained as a sample but as an experience.

Do we want to track the cloud-run-hello or fork? If we want to fork, we should consider simplifying. For example, instead of the full web page currently generated, we could go with:

text/plain:
```
Service ??? running Revision ??? in Project ???
```

I asked this question because I don't have a strong sense of the value of the more visual experience in this case.

golang/go-cloud-run-hello-world/README.md
grayside: This is copied from other sample READMEs, I'll iterate this for Cloud Run, but maybe there are changes to make to the existing sample READMEs?

golang/go-cloud-run-hello-world/README.md
grayside: Alright, will iterate this. However, I don't think it's good to hide the command-line operations. It might make sense for a simple example like this, but once it's more complicated understanding the escape valve is important.

golang/go-cloud-run-hello-world/cloudbuild.yaml
grayside: Good callout! The current best approach is to avoid duplication by naming the service distinctly, examples would be something like:

* hello-world-$SHORT_SHA
* hello-world-$BUILD_ID

If there can be a build that runs multiple samples, we can re-introduce the language:

* hello-world-go-$SHORT_SHA
* hello-world-go-$BUILD_ID
* hello-world-go-$_PR_NUMBER

PR Number would require some extra finesse if there's anything like a nightly build of all samples.

Happy to discuss future options but this is the best practice for now. Do you have a preference?

golang/go-cloud-run-hello-world/main.go
grayside: Current behavior is wait for timeout then it falls back to default `???`.
We could preempt the metadata server request by checking for a `GOOGLE_CLOUD_PROJECT` env var. Getting the Project ID on Cloud Run has a metadata server dependency.

golang/go-cloud-run-hello-world/main.go
grayside: This isn't required to run locally. It's just nice to have in terminal output, whereas it's a distraction in stackdriver.

Options:
1. Remove and always have timestamps
2. Leave in place and let developers edit the code if they want to.
3. Add a check for something like `DEV=1` env var and toggle accordingly.

The plan is for Cloud Code to simulate all the mechanisms we might use to detect running in Cloud Run.

golang/go-cloud-run-hello-world/main.go
grayside: 8080 is the default and the port is output as part of server start. That seems sufficient to me unless the goal is to document the availability of the PORT var to avoid port conflicts? If so, how does that relate to only using Cloud Code command palette to run locally?

golang/go-cloud-run-hello-world/.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Run on Cloud Run",
+      "type": "cloudcode.kubernetes",
```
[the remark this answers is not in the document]
grayside: Done

golang/go-cloud-run-hello-world/README.md
grayside: Mostly done. I don't think all the useful links exist yet.

golang/go-cloud-run-hello-world/README.md
grayside: Made some structural changes to try covering both IDEs more even-handedly. There's gaps for IntelliJ still.

golang/go-cloud-run-hello-world/README.md
grayside: Removed this given transparency and determination not to include CLI steps.

grayside:
I've complete a feedback response round. Updates are mostly README revisions, otherwise there are several conversation threads exploring further changes/requirements.

golang/go-cloud-run-hello-world/index.html
simonz130: The value of more visual experience is giving developers better sense of accomplishment. It also demoes better compared to a plain text.
I agree, we should fork and simplify where appropriate, we don't need a dark mode or other more advanced UI-related scenarios or features.

golang/go-cloud-run-hello-world/cloudbuild.yaml
simonz130: <lang>-hello-world-cloud-run-$short_sha 
The same with $build_id sounds good too. Up to you.

I'm not familiar with Cloud Run networking, so what happens when you want to expose your Cloud Run service to internet? Will its IP object have a name that you can manage, like service/load balancer in kubernetes? Or is it internally managed by Cloud Run?
Asking to make sure we avoid conflicts in that too.

golang/go-cloud-run-hello-world/main.go
simonz130: I like that!

Check the GOOGLE_CLOUD_PROJECT env var, if it's set, wait for timeout and fall back if needed. If it's not set, it means you run locally.

golang/go-cloud-run-hello-world/main.go
simonz130: Could we generalize option 3 a little bit? So instread of introducing a new environment variable for local vs. cloud, we can test the same logic you already have for determining whether to call metadata server.

golang/go-cloud-run-hello-world/main.go
simonz130: In case developers run multiple workloads in containers locally, we may want to let them know that this sample is using $PORT to expose the service locally.

golang/go-cloud-run-hello-world/cloudbuild.yaml
grayside: No need to worry about ip addresses, that's completely abstracted by the platform.

golang/go-cloud-run-hello-world/main.go
grayside: Leaning on GOOGLE_CLOUD_PROJECT env var for this. It's a bit awkward for conceptual flow but adaptive is better than commenting instructions.

golang/go-cloud-run-hello-world/main.go
grayside: Done.

golang/go-cloud-run-hello-world/cloudbuild.yaml
grayside: Ended up going with COMMIT_SHA to avoid multiple substitutions. Will switch the whole thing to SHORT_SHA if something proves too long.

golang/go-cloud-run-hello-world/main.go
grayside: Added README section for `PORT` and `GOOGLE_CLOUD_PROJECT`

grayside:
Completed last round on this PR for today. Thanks for taking the time to go around twice!

grayside:
Changes in recent updates:

* Updated Congrats => Congratulations
* Refactored integration test to a go implementation
* Added local service CLI testing instructions to README

PR #228, opened 2020-02-25
Python cloud run hello with Flask

engelke:


python/cloud-run-python-hello-world/.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/python-hello-world.iml" filepath="$PROJECT_DIR$/.idea/python-hello-world.iml" />
```
simonz130: This needs to be adjusted in all relevant places to be _cloud-run-python-hello-world_ instead of python-hello-world.

python/cloud-run-python-hello-world/.idea/python-hello-world.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
```
simonz130: Rename the file to cloud-run-python-hello-world.iml

python/cloud-run-python-hello-world/.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
...
+      "portForward": true
+    },
+    {
+      "name": "Debug on Kubernetes",
+      "type": "cloudcode.kubernetes",
+      "request": "attach",
+      "podSelector": {
+        "app": "python-hello-world"
+      },
+      "localRoot": "${workspaceFolder}/src",
```
simonz130: A heads up that we'll need to make changes in launch.json to accomodate debugging Cloud Run container on minikube. We don't have the details yet but this will definitely change.

python/cloud-run-python-hello-world/app.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_metadata(item_name):
+    metadata_url = 'http://metadata.google.internal/computeMetadata/v1/'
+    headers = {'Metadata-Flavor': 'Google'}
+
+    try:
+        r = requests.get(metadata_url + item_name, headers=headers)
+        return r.text
+    except:
+        return 'Unavailable'
```
simonz130: Could we refine it so that it distinguishes between running locally vs running on Cloud Run and not being able to detect the project ID?
If it runs locally, we should either print the project ID set in local gcloud config or at least print a note that makes it clear that it's not available due to running locally.

python/cloud-run-python-hello-world/app.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@app.route('/')
+def hello():
+    """Return a friendly HTTP greeting."""
+    message = "Hello World"
+
+    project = get_metadata('project/project-id')
+    service = os.environ.get('K_SERVICE', 'Unknown service')
+    revision = os.environ.get('K_REVISION', 'Unknown revision')
```
simonz130: Will this work the same when running on a local minikube vs. on Cloud Run?

python/cloud-run-python-hello-world/app.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return render_template('index.html',
+        message=message,
+        Project=project,
+        Service=service,
+        Revision=revision)
+
+if __name__ == '__main__':
+    server_port = os.environ.get('PORT')
```
simonz130: could you fall back to port 8080 if environment variable of port is not set?

python/cloud-run-python-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,30 @@
...
+  args: ['build', '-t', 'gcr.io/${_PROJECT_ID}/${_SERVICE}', '.']
+
+- id: docker push
+  name: 'gcr.io/cloud-builders/docker'
+  args: ['push', 'gcr.io/${_PROJECT_ID}/${_SERVICE}']
+
+- id: Deploy to Cloud Run
+  name: 'gcr.io/cloud-builders/gcloud'
+  args:
+  - beta
```
simonz130: is it still beta?
I was able to deploy using **gcloud run deploy**

python/cloud-run-python-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - run
+  - deploy
+  - ${_SERVICE}
```
simonz130: This might be an issue in case multiple CI pipelines will run in parallel - they might collision.
We need a way to have each CI run create a separate service that can be tested and deleted without impacting other CI runs.
Look at the similar discussion we had with Adam in Go PR.

python/cloud-run-python-hello-world/.idea/modules.xml
engelke: Done in next commit.

python/cloud-run-python-hello-world/.idea/python-hello-world.iml
engelke: Done in next commit.

python/cloud-run-python-hello-world/app.py
engelke: Done in next commit.

python/cloud-run-python-hello-world/cloudbuild.yaml
engelke: Done in next commit.

python/cloud-run-python-hello-world/cloudbuild.yaml
engelke: Done in next commit.

python/cloud-run-python-hello-world/app.py
engelke: TBD - don't know yet

python/cloud-run-python-hello-world/app.py
grayside: Let's standardize on using GOOGLE_CLOUD_PROJECT env var to preempt the use of metadata server. Not quite local detection but meaningful as an override. The Go PR includes this.

python/cloud-run-python-hello-world/app.py
grayside: Note the canonical sample fallback is `???`, not `Unknown service`.
In local we expect to see `???` in the output unless something explicitly sets these env vars. Cloud Code's local emulation is expected to do something along these lines eventually if I understood correctly.

python/cloud-run-python-hello-world/app.py
grayside: Please compare with https://cloud.google.com/run/docs/tutorials/pubsub for existing practices & syntax for Cloud Run Python.

PR #230, opened 2020-02-27
Dotnet cloud run

simonz130:


simonz130:
@grayside, PTAL

.templates/template_location.json
```diff
@@ -4,83 +4,103 @@
     "directoryPath": "python/python-hello-world",
     "templatePath": "python-hello-world",
     "templateName": "Python (Flask): Hello World",
-    "templateLanguages": ["python"]
+    "templateLanguages": ["python"],
+    "run-platform": ["kubernetes"]
```
peterlin741: Just a suggestion (since we're consuming this for the IntelliJ plugin), you can rename this field from `run-platform` to `runPlatforms` (plural for platforms since this is a list, and removing the dash, which may cause issues when serializing/deserializing)

etanshaul: +1. and regardless of consuming platform, the format should be consistent across these fields.

also +1 to making it plural since its an array type

dotnet/dotnet-cloud-run-hello-world/.vscode/tasks.json
```diff
@@ -0,0 +1,36 @@
+{
```
grayside: Is this file meant to be generated? It's unclear to me what we want to include here.

dotnet/dotnet-cloud-run-hello-world/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /src
+COPY . .
+RUN dotnet restore dotnet-cloud-run-hello-world.csproj
+RUN dotnet build "./dotnet-cloud-run-hello-world.csproj" -c Debug -o /out
+
+FROM build AS publish
+RUN dotnet publish dotnet-cloud-run-hello-world.csproj -c Debug -o /out
+
+# Building final image used in running container
+FROM base AS final
```
grayside: The distinct declaration of base which is not used otherwise seems unnecessary. If we're going to include it as a matter of training developers on multi-stage caching strategies, we should include a comment explaining it.

Should we add this to Java & Go?

dotnet/dotnet-cloud-run-hello-world/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /src
```
grayside: Redundant to base image.

dotnet/dotnet-cloud-run-hello-world/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY --from=publish /out .
+ENV ASPNETCORE_URLS=http://*:8080
```
grayside: Hard-coding 8080 is not compliant with the runtime contract. We want the service to bind on $PORT environment variable at runtime. If we must specify 8080 here, we should also deploy with `--port 8080`, but that's really intended to support services developers cannot modify.

dotnet/dotnet-cloud-run-hello-world/Startup.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                app.UseDeveloperExceptionPage();
+            }
+            else
+            {
+                // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
+                app.UseHsts();
+            }
+
+            app.UseHttpsRedirection();
```
grayside: Cloud Run itself redirects to HTTPS, some of these options may be redundant, but good to keep with a comment for portability.

.templates/template_location.json
simonz130: np, changing it to runPlatforms

dotnet/dotnet-cloud-run-hello-world/.vscode/tasks.json
simonz130: Yes, it should but the content depends on the language.
For .net the goal was to have tasks for building and packaging of the code locally. It can contain other tasks that are common, such as running unit tests, if it's applicable.

dotnet/dotnet-cloud-run-hello-world/Dockerfile
simonz130: Yes, I think we should add it to other languages too and good point on the comment. I'll add it.

dotnet/dotnet-cloud-run-hello-world/Dockerfile
simonz130: Very good point, thanks! It should have the port specified via $PORT.

dotnet/dotnet-cloud-run-hello-world/Startup.cs
simonz130: Great point, added a comment. Thanks!

PR #233, opened 2020-03-03
Add Cloud Run (nodejs)

grayside:
Add the Cloud Run Hello World sample for Node.js.

This tracks #217 for language-agnostic implementation details beyond the design.

## Interesting Bits

* Includes eslint for code style checks (npm run lint; npm run fix) and adds lint as the first step of the CI pipeline.
* Using mocha to drive integration tests, configured so `npm test` will also work locally.
* Handlebars for templating

## TODOS

* [ ] Node.js Code Review
* [ ] Confirm linting rules
* [ ] Confirm inclusion of code style as a CI step.
* [ ] Finalize sample design
* [ ] Complete Cloud Code integration
* [ ] Complete README


nodejs/nodejs-cloud-run-hello-world/app.js
```diff
@@ -0,0 +1,51 @@
...
+  project: process.env.GOOGLE_CLOUD_PROJECT,
+  service: process.env.K_SERVICE || '???',
+  revision: process.env.K_REVISION || '???',
+};
+let template;
+
+app.get('/', async (req, res) => {
+  // Prepare the template for use if not ready.
+  if (!template) {
+    // Only check the Cloud Run metadata server if it looks like the service
```
engelke: Re: this comment. How does this check to see if it looks like the service is deployed to Cloud Run. Doesn't it just check for the environment variable?

nodejs/nodejs-cloud-run-hello-world/index.html.hbs
```diff
@@ -0,0 +1,176 @@
...
+			<div style="text-align:center;">
+				<picture>
+					<source srcset="/assets/celebration-dark.svg" alt="A group celebrating" media="(prefers-color-scheme: dark)">
+					<img src="/assets/celebration.svg" alt="A group celebrating" width="427" height="231">
+				</picture>
+			</div>
+
+			<div class="message">
+				<h1>It's running!</h1>
+				<h2>Congrats, you successfully deployed a container image to Cloud Run</h2>
```
engelke: Tiny nit - non-native English speakers might not recognize 'Congrats' as 'Congratulations'.

nodejs/nodejs-cloud-run-hello-world/app.js
averikitsch: Cloud Run doesn't have the project id as an environment variable. This comment though is a little confusing, specifically around the phrase "if it looks like the service...".

grayside: The idea is we check the metadata server if we don't have the environment variable, but the environment variable is only set if the developer explicitly sets it.

I softened the comment language because neither of these are definitive about being on Cloud Run or being local, will try to clarify it.

nodejs/nodejs-cloud-run-hello-world/index.html.hbs
grayside: We get "Congrats" from the source template in cloud-run-hello, but this is a a good callout. Let's change to Congratulations in all the samples. The frontend experience should be identical across all of them.

grayside:
In the latest commits, this happened:

* Added comments and clarified existing comments.
* Reflowed the one-time code inside the handler so the check for template is followed by loading the template, instead of first by loading the Project ID.
* Tweaked the README which currently has a local testing CLI command in addition to instructions for manual Cloud Build submit.

nodejs/nodejs-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+This sample demonstrates how to use the Cloud Code extension in VS Code.
+
+* [Install Cloud Code for VS Code](https://cloud.google.com/code/docs/vscode/install)
+* [Creating a new app](https://cloud.google.com/code/docs/vscode/creating-an-application)
+* [Editing YAML files](https://cloud.google.com/code/docs/vscode/yaml-editing)
+
+## Running locally
+
+1. Open the command pallette
```
averikitsch:
```suggestion
1. Open the command palette
```

nodejs/nodejs-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,87 @@
...
+This sample demonstrates how to use the Cloud Code extension in IntelliJ.
+
+* [Install Cloud Code for IntelliJ](https://cloud.google.com/code/docs/intellij/install)
+* [Creating a new app](https://cloud.google.com/code/docs/intellij/create-run-app)
+
+## Service Configuration
+
+Configuration for this service uses environment variables.
+
+* **`GOOGLE_CLOUD_PROJECT`** [default: `<none>`] Override for the Project ID. If set the service assumes it's running locally and does not use the metadata server.
```
simonz130: why is this needed?

nodejs/nodejs-cloud-run-hello-world/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Next Steps
+
+* Read the Cloud Run documentation on [developing your service](https://cloud.google.com/run/docs/developing).
+* Follow the [System packages tutorial](https://cloud.google.com/run/docs/tutorials/system-packages) to learn how to use the command-line to build and deploy a more complicated service.
+* {Cloud Code link?}
+
+## Contributing
+
+Join us at https://github.com/GoogleCloudPlatform/cloud-code-samples.
```
simonz130: Let's remove this for now.

nodejs/nodejs-cloud-run-hello-world/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Use the official lightweight Node.js 12 image.
+# https://hub.docker.com/_/node
+FROM node:12-slim
+
+# Create and change to the app directory.
+WORKDIR /usr/src/app
+
+# Copy application dependency manifests to the container image.
```
simonz130: In other apps we are copying the source and building final container image - all in docker file, so it's self contained.

nodejs/nodejs-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+- id: 'Code Style'
+  name: 'node:12-slim'
+  entrypoint: '/bin/bash'
+  args:
+  - '-c'
+  - |
+    npm install
+    npm run lint
+
```
simonz130: Why do you need this step? Building docker image based on Dockerfile should handle this.

nodejs/nodejs-cloud-run-hello-world/cloudbuild.yaml
```diff
@@ -0,0 +1,76 @@
...
+- id: 'Build Container Image'
+  name: 'gcr.io/cloud-builders/docker:latest'
+  args: ['build', '-t', 'gcr.io/${PROJECT_ID}/${_SERVICE_TAG}:${COMMIT_SHA}', '.']
+
+- id: 'Push Container Image'
+  name: 'gcr.io/cloud-builders/docker:latest'
+  args: ['push', 'gcr.io/${PROJECT_ID}/${_SERVICE_TAG}:${COMMIT_SHA}']
```
simonz130: In .net template I changed it to ${_SERVICE_TAG}${COMMIT_SHA} (removed : )

The reason: when invoking cloud build locally using cloudbuild.yaml, it errors due to special characters. $COMMIT_SHA is null when invoking locally. I suggest we do that in the rest of the templates so they are easily invokable both in CI pipelines and locally.

nodejs/nodejs-cloud-run-hello-world/README.md
grayside: This was added per request in the golang sample: it documents the environment variable that specifies the project, and prevents the need to ping the metadata server. This architecture & documentation pattern should be used by all samples unless we choose to change the app design.

nodejs/nodejs-cloud-run-hello-world/README.md
grayside: It would be helpful to provide feedback like this on the Go PR, as this comes from the template there. Will remove in both places.

nodejs/nodejs-cloud-run-hello-world/Dockerfile
grayside: That's what's happening here.

nodejs/nodejs-cloud-run-hello-world/cloudbuild.yaml
grayside: I switched to running tests via JS. That requires dev dependencies. We don't install them inside the container currently, we could switch to running them inside but then the container is further from production ready.

I can remove this "Code Style" step, I added it because I had convenient linting scripts in the sample and needed the node_modules installed for the integration step.

nodejs/nodejs-cloud-run-hello-world/cloudbuild.yaml
grayside: My solution was to keep a "normal" tag approach and document how to run the pipeline:

```
gcloud builds submit . --config cloudbuild.yaml --substitutions COMMIT_SHA=manual
```

I don't think this is much more arcane to the typical developer and manually running the tests won't be something developers typically do with the code.

simonz130: I like that. Will change the dotnet one accordingly.

nodejs/nodejs-cloud-run-hello-world/.idea/externalDependencies.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+    <component name="ExternalDependencies">
```
bcoe: probably we don't want to check in this `.idea` folder?

nodejs/nodejs-cloud-run-hello-world/app.js
```diff
@@ -0,0 +1,54 @@
...
+  revision: process.env.K_REVISION || '???',
+};
+let template;
+
+app.get('/', async (req, res) => {
+  // The handlebars template is stored in global state so this will only once.
+  if (!template) {
+    // Load Handlebars template from filesystem and compile for use.
+    try {
+      const source = await fs.readFile('index.html.hbs', 'utf-8');
```
bcoe: nit, but since this resource is only needs to be loaded once, I might just:

```js
const template = handlebars.compile(fs.readFileSync('index.html.hbs', 'utf8'));
```

eliminates a little bit of boilerplate, if this is your only fs access, you could then:

```
const {readFileSync} = require('fs');
```

rather than pulling in `fs.promises`.

nodejs/nodejs-cloud-run-hello-world/index.html.hbs
```diff
@@ -0,0 +1,176 @@
+<!doctype html>
+<html lang=en>
+<head>
+<meta charset=utf-8>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>Congratulations | Cloud Run</title>
+<link href="https://fonts.googleapis.com/css?family=Roboto" rel="preload" as="font">
+<link
+  href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAYAAABzenr0AAAACXBIWXMAAAsSAAALEgHS3X78AAAB5klEQVRYha2XMVLDMBBFPxl6qCjciBskNzA3AE5AqrTQuaVVR1pX4QSYGyQ3SGqGATcu3JBUlGGUiTPyeiWvZH6TsSTLb/+XN8nZfr+H0lUBYAvgucySbzBSupoCmJRZ8sTNx2p0vG8J4AHAl9LVQunqmtnPQD4qXa0d81GyARo1IC/2hmWWGIfeAYwBGIi7fwMos2QNYEfmTLWXZKw4fl4AeKOQQxwAcaERrbIg14MjCQKwYrA1KJI+gBtmjLqAIZEcXsNGSlfb42a27sssKaw15lz8ePbcGOdcrzPViFz3uuCIwVZQJBIAbiMuBlviSCQASulqEgjQqPctaQE4+gEiYrDljYQ6AIcLU2ZM6gJ8kUgBxp6uGKJOJFIACJuSRK1IOgCecxDzNrh0iuScW/D7+eG4LxGuk6sDcDVbTZhuCEe1sV/JxuFpnacF5wDX/zsAV7OVOZS3EQ8/tOo6Tw+tWgqwqfN0S8Ziqp/Xedr6SScFWDBjIQAny+lEC8CT/5KsC7G/ZbkXwFF9WefpmoxJq+9YHgMQc/qdlscAhNrvtdwJ4Mh/x1Thq77XcieApHoPgNjyUABJ8wmyfBAAU32w5SyAI/+5p/tFW84CkOpfzb9kaqll/yDLfQDsgy2Z6gdb3hKAP2n+6m+fxBVHAAAAAElFTkSuQmCC"
```
bcoe: what's this image? a comment might be worthwhile, or it might even be worth loading the image from an asset.

nodejs/nodejs-cloud-run-hello-world/index.html.hbs
```diff
@@ -0,0 +1,176 @@
...
+			<div class="message">
+				<h1>It's running!</h1>
+				<h2>Congratulations, you successfully deployed a container image to Cloud Run</h2>
+			</div>
+		</div>
+
+		<div class="details">
+			<p>
+                This created the revision <code>{{revision}}</code> of the Cloud Run service <code>{{service}}</code>
+                {{#if project}} in the GCP project <code>{{project}}</code>{{/if}}.
```
bcoe: some of the layout looks slightly weird, I think you could setup prettier to check the HTML:

https://github.com/GoogleCloudPlatform/wombat-dressing-room/blob/master/package.json#L13

nodejs/nodejs-cloud-run-hello-world/index.js
```diff
@@ -0,0 +1,11 @@
+const app = require('./app');
+const pkg = require('./package');
+const PORT = process.env.PORT || 8080;
+
+app.listen(PORT, () => {
+  console.log(
+    'Hello from Cloud Run! The container started successfully and is listening for HTTP requests on $PORT'
```
bcoe: I've tended to have the `.listen` in the same file as the `app`, or inverted things a bit, where `app` would expose methods, and the harness would just wire up the routes.

I can see the advantage to this approach, if you intend to add a couple additional routes on to app; curious why you like this design approach?

nodejs/nodejs-cloud-run-hello-world/test/system.test.js
```diff
@@ -0,0 +1,20 @@
+const assert = require('assert');
+const {request} = require('gaxios');
+
+const port = process.env.PORT || '8080';
+const url = process.env.SERVICE_URL || `http://localhost:${port}`;
+
+describe('Hello World', () => {
+  it('can respond to an HTTP request', async () => {
+    console.log(`    - Requesting GET ${url}/...`)
```
bcoe: what I used to do, which I thought worked really well, would be to actually stand up and tear down the web application in the `before` statement of the test; I believe you might be able to bind to `0` (maybe?) and get a random port, and there should be a method `express` exposes for terminating the server.

nodejs/nodejs-cloud-run-hello-world/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "nodejs-cloud-run-hello-world",
+  "version": "0.0.1",
+  "description": "Hello World application for Cloud Run",
+  "main": "index.js",
+  "scripts": {
+    "start": "node index.js",
+    "debug": "node --inspect=9229 index.js",
+    "lint": "eslint **.js",
+    "fix": "eslint --fix *.js",
+    "test": "mocha test/*.test.js --check-leaks --timeout 5000"
```
bcoe: If you replaced this with:

```
c8 mocha test/*.test.js --check-leaks --timeout 5000
```

and `npm i c8 --save-dev` you'll get test coverage for free, your call on whether you want to do so :smile:

nodejs/nodejs-cloud-run-hello-world/.idea/externalDependencies.xml
grayside: This is a requirement of the sample, though the contents of this folder may change.

nodejs/nodejs-cloud-run-hello-world/index.js
grayside: It's a legacy of using supertest, where injecting requests into the app object is part of the API. I'd probably do as you say and and have app.js just expose methods, but from a sample standpoint I think it confuses reading individual slices of code to aggressively separating routing from logic.

nodejs/nodejs-cloud-run-hello-world/test/system.test.js
grayside: In this case, I'm testing against the service running on Cloud Run, and for reasons of 1. Developer Education, 2. Lack of a graceful way to handle Cloud Run deploys the setup/teardown is externalized. I have ideas around a hand-written library in the future happy to discuss

nodejs/nodejs-cloud-run-hello-world/index.js
grayside: I'll go ahead and combine them into one file.

PR #247, opened 2020-03-19
Update CI for Java Cloud Run

averikitsch:


java/CONTRIBUTING.md
```diff
@@ -0,0 +1,32 @@
...
+Join us at https://github.com/GoogleCloudPlatform/cloud-code-samples.
+
+## Run the Tests
+
+Some of the samples are tested by a dedicated [Cloud Build](https://cloud.google.com/cloud-build)  configuration.
+
+All sample tests are expected to take the following steps:
+
+* Build the service
+* Deploy to Cloud Run
```
simonz130: We should generalize here. These instructions should serve not only Cloud Run but also kubernetes.
Adam has the same remark. We'll merge his updates as a canonical and then update across all the samples, so no action item on this yet, just heads up.

PR #252, opened 2020-03-25
Consistent use of Go language identifier in the template metadata

etanshaul:
The k8s sample use "go". The CR samples use "golang". The extensions hinge on this field for determining which templates are appropriate to show. we should be consistent.


etanshaul:
@simonz130 @grayside ptal

simonz130 (approved):

PR #257, opened 2020-03-31
[Cloud Run] Update ReadMes

averikitsch:


java/java-cloud-run-hello-world/Dockerfile
```diff
@@ -1,25 +0,0 @@
-# Use the official maven/Java 8 image to create a build artifact.
```
grayside: Was the Java Dockerfile meant to be removed in this PR?

averikitsch: I should move it to another PR. I need to talk to the team about mixing Dockerfiles and Jib. We should probably choose one or the other.

PR #259, opened 2020-03-31
Add environment variable to better support skaffold debug

etanshaul:
`skaffold debug` uses heuristics to determine if an application can be debugged, as described here:
https://skaffold.dev/docs/workflows/debug/

In it's current state, the [Go CR hello-world app](https://github.com/GoogleCloudPlatform/cloud-code-samples/tree/cloudrun-templates/golang) doesn't work with `skaffold debug` without one of the standard go-runtime env vars as explained in the docs.

### DIFF
--- a/golang/go-cloud-run-hello-world/Dockerfile
+++ b/golang/go-cloud-run-hello-world/Dockerfile
@@ -20,6 +20,9 @@ RUN go build -mod=readonly -v -o /app
 COPY index.html /index.html
 COPY assets ./assets/
 
+# Add stack traces for all user-created goroutines
+ENV GOTRACEBACK=all
+
 # If you want to use the debugger, you need to modify the entrypoint to the
 # container and point it to the "dlv debug" command:
 #   * UNCOMMENT the following ENTRYPOINT statement,
--- a/golang/go-cloud-run-hello-world/Dockerfile
+++ b/golang/go-cloud-run-hello-world/Dockerfile
@@ -20,8 +20,8 @@ RUN go build -mod=readonly -v -o /app
 COPY index.html /index.html
 COPY assets ./assets/
 
-# Add stack traces for all user-created goroutines
-ENV GOTRACEBACK=all
+# Default behavior - a failure prints a stack trace for the current goroutine
+ENV GOTRACEBACK=single
 
 # If you want to use the debugger, you need to modify the entrypoint to the
 # container and point it to the "dlv debug" command:
--- a/golang/go-cloud-run-hello-world/Dockerfile
+++ b/golang/go-cloud-run-hello-world/Dockerfile
@@ -20,7 +20,9 @@ RUN go build -mod=readonly -v -o /app
 COPY index.html /index.html
 COPY assets ./assets/
 
-# Default behavior - a failure prints a stack trace for the current goroutine
+# Definition of this variable is used by 'skaffold debug' to identify a golang binary.
+# Default behavior - a failure prints a stack trace for the current goroutine.
+# See https://golang.org/pkg/runtime/
 ENV GOTRACEBACK=single
 
 # If you want to use the debugger, you need to modify the entrypoint to the


etanshaul:
@briandealwis fyi for review as well

etanshaul:
@simonz130 @averikitsch @grayside ptal. Thanks.

etanshaul:
@briandealwis do you have any more insight into @grayside 's questions? My understanding is that this is a commonly used Go setting.

Regarding remove `dlv`. Yes, IntelliJ doesn't need this. In fact, I'd love it if debug config from ALL of our samples (k8s) included could be removed. But we need to confirm first with VS Code: @sujit-kamireddy , @quoctruong please for comment regarding removal of debug config from the Dockerfile(s).

briandealwis:
There’s a chicken-and-egg problem: Go programs can’t be identified from examining the container image metadata alone. So we can’t inject the variable. The [`GOTRACEBACK` docs](https://golang.org/pkg/runtime/?m=all) doesn’t describe any performance impacts, but it’s important to note that it can be set to any value, including `none`.  (The default value is `single`.)

The stack traces are emitted stderr, and so should be picked up by the container logging.

etanshaul:
@briandealwis so if it is just set to the default value of `none`, that is sufficient for `skaffold debug`'s heuristics right?

@grayside if you are concerned about other implications of adding this env var, then for now we can just set it to the default value (which is basically a noop).

etanshaul:
(I updated the PR to use `single`)

grayside:
I'm not as concerned about performance as I am about making sure we're communicating why we do things that are not standard practice.

Is this setting needed for debug or a proxy to identify that we're running a go binary? If the latter, wouldn't it make more sense to define a CLOUD_CODE_RUNTIME override variable for go or other edge cases?

etanshaul:
> Is this setting needed for debug or a proxy to identify that we're running a go binary? 

Yes, this is needed to identify that we are running a go binary and that we can debug it.

> If the latter, wouldn't it make sense to define a dedicated CLOUD_CODE_RUNTIME override variable for go or other edge cases?

The hope is that we'd be able to debug most applications in their default state _without having to add any extra stuff (like that override that you mention)_ . I can see how you may think though that we are doing the same thing by adding the `GOTRACEBACK` env var (even though it is a standard thing to have). @briandealwis is there nothing in the base `goland` image already configured that Skaffold can add to its heuristics?

grayside:
It's pretty common for a golang Dockerfile to be multi-stage, copying the binary into an image without the go toolchain. Would skaffold debug work in that case if it can identify the binary is built by go?

briandealwis:
We can't identify that a binary is a Go binary.  We've considered introducing explicit means to configure the language runtime type (GoogleContainerTools/skaffold#2203) but this is somewhat intrusive.  And as it's [not](https://medium.com/honestbee-tw-engineer/how-did-we-fine-tune-the-go-application-performance-76f3a6f77736) [uncommon](https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap-26c2462549a2/) to use one of several Go runtime-specific environment variable(s), particularly `GOGC`, it makes sense to use them.

etanshaul:
@simonz130 @grayside could we try to come to a resolution on this? It is blocking debugging support on our Cloud Run Go sample

briandealwis:
I realized I didn't address some of @grayside's questions/comments.  Just to be clear:

> Does this mean the dlv tool can be removed?

Yes, providing we identify the image as being Go related via one of these environment variables (more below).

> While we can add the variable to the Dockerfile, it seems worth asking: should we inject this into the container when starting locally so there's one less thing specializing the Dockerfile for development vs. production?

This is great idea.  `skaffold debug` can look at the source and try to identify the source types.  I've opened issue an issue (GoogleContainerTools/skaffold#3977).  This will require some time to implement.

> Are their performance implications to this variable? Haven't played with this one much. If we want to guide developers to remove this when not debugging I can follow-up with commented instructions.

There are no performance implications: the use of `single` is the default value used by the Go runtime.

> Is this setting needed for debug or a proxy to identify that we're running a go binary? If the latter, wouldn't it make sense to define a dedicated CLOUD_CODE_RUNTIME override variable for go or other edge cases?

Yes, these variables serve as a proxy identifier.  It's not uncommon for containers to define these values, particularly `GOGC` or `GOTRACEBACK=all` (if you had a crash in a binary, you'd want to see everything that was going on).

Settling on an agreeable prefix is … difficult.  This functionality is delivered in Skaffold and is used independently of Cloud Code.  It's actually why I settled on using these proxy identifiers instead as they existed and were in use.

> It's pretty common for a golang Dockerfile to be multi-stage, copying the binary into an image without the go toolchain. Would skaffold debug work in that case if it can identify the binary is built by go?

It works for multistage builds: `debug` currently examines the final image that's produced.(Providing that the environment variable is defined in the final stage!)

golang/go-cloud-run-hello-world/Dockerfile
```diff
@@ -20,6 +20,9 @@ RUN go build -mod=readonly -v -o /app
 COPY index.html /index.html
 COPY assets ./assets/
 
+# Default behavior - a failure prints a stack trace for the current goroutine
```
grayside, line 23:
```suggestion
# Definition of this variable is used by 'skaffold debug' to identify a golang binary.
# Default behavior - a failure prints a stack trace for the current goroutine.
# See https://golang.org/pkg/runtime/
```

etanshaul:
thanks @grayside . Change suggestion accepted. ptal.

PR #280, opened 2020-04-08
Cloudrun templates

simonz130:
- Adding new templates json that includes Cloud Run templates
- Reverting previously used templates json to the list of kubernetes templates only.

This will allow one extension to release ahead of the other one.


etanshaul:
@peterlin741 fyi

PR #310, opened 2020-05-07
Fixing Python and Go Cloud Run CI

simonz130:
We add a working dir so Cloud Build CI is not confused when invoking trigger. There is currently no way to specify the working dir for a trigger globally.


simonz130:
Merging this as it improves the situation, but we'll need to update the generated name of the Cloud Run service due to a Cloud Run bug with name's length <=63 chars.

Issue #305, opened 2020-04-29
[BUG] CI is not triggered on cloud run samples


simonz130:
**Describe the bug**
PRs submitted in folders with Cloud Run samples are not triggering a Cloud Build trigger.

**To Reproduce**

**Expected behavior**

**Screenshots**

** Template in use:**
 - Language
 - Path to template

**Cloud Code in use:**
 - [Cloud Code for VS Code] | [Cloud Code for IntelliJ]

**Additional context**

PR #311, opened 2020-05-08
Vscode run readmes

simonz130:
- Added vscode instructions to all Cloud Run samples in readmes.
- Updated template HTML to point to "next steps" that include both vscode and intellij now.


simonz130:
The CI is failing due to #305, we should ignore CI failures for this PR. The fix for CI is worked out in parallel.

PR #318, opened 2020-05-11
Increase timeout waiting for MongoDB

vincentjocodes:
When deploying to GKE using Cloud Code, sometimes connection to DB takes longer than 10 seconds so it fails the backend server for debugging.

The error says "ping to mongodb failed: context deadline exceeded". After increasing the timeout, I haven't seen this error.

### DIFF
--- a/golang/go-guestbook/src/backend/main.go
+++ b/golang/go-guestbook/src/backend/main.go
@@ -42,7 +42,7 @@ func main() {
 	}
 
 	mongoURI := "mongodb://" + dbAddr
-	connCtx, cancel := context.WithTimeout(ctx, time.Second*10)
+	connCtx, cancel := context.WithTimeout(ctx, time.Second*30)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


simonz130:
I see you are fixing it in Go sample. Is it happening only in Go? What about rest of the languages?

vincentjocodes:
> I see you are fixing it in Go sample. Is it happening only in Go? What about rest of the languages?

Hi Simon, I haven't encountered this problem in other samples as far as I remember.

sujit-kamireddy:
> > I see you are fixing it in Go sample. Is it happening only in Go? What about rest of the languages?
> 
> Hi Simon, I haven't encountered this problem in other samples as far as I remember.

I wonder why is it prevalent in just Go. I would fix it in all samples to be consistent.

simonz130:
It's possible that GO driver for Mongo is less performant than in other languages? Not sure if this is true but I'll merge this change and then if we encounter an issue in the future, we'll update the rest of them as appropriate.

PR #319, opened 2020-05-12
Add environment variable to allow connection

vincentjocodes:
If this is not set, the DB container won't start up. Causes users to not be able to Deploy or to Debug the containers.

Errors I get from GKE container logs:

image

Relevant PR:
https://github.com/docker-library/postgres/pull/658

### DIFF
--- a/python/django/python-guestbook/kubernetes-manifests/guestbook-postgres.deployment.yaml
+++ b/python/django/python-guestbook/kubernetes-manifests/guestbook-postgres.deployment.yaml
@@ -28,3 +28,6 @@ spec:
           name: db
           ports:
             - containerPort: 5432
+          env:
+            - name: POSTGRES_HOST_AUTH_METHOD
+              value: "trust"


python/django/python-guestbook/kubernetes-manifests/guestbook-postgres.deployment.yaml
```diff
@@ -28,3 +28,6 @@ spec:
           name: db
           ports:
             - containerPort: 5432
+          env:
+            - name: POSTGRES_HOST_AUTH_METHOD
+              value: "trust"
```
simonz130, line 33: Don't we need the same to be set on hello-world sample for Python (django version)?

vincentjocodes: We don't since the hello world Django example doesn't have a Postgres deployment. Thanks for checking.

PR #364, opened 2020-06-26
Make adjustments to Python projects for Skaffold buildpack support

j-windsor:
Add procfiles and prevent wheel building of psycopg2 to allow Skaffold to detect and build with Buildpacks successfully.

Note that we are not fixing python django guestbook here. That is currently blocked by https://github.com/GoogleContainerTools/skaffold/issues/4397


etanshaul:
@j-windsor looks like we may have missed the k8s Guestbook _django_ sample?

fsladkey:
I think that one's blocked.

> Note that we are not fixing python django guestbook here. That is currently blocked by GoogleContainerTools/skaffold#4397

etanshaul:
oh. thanks @fsladkey.

PR #361, opened 2020-06-22
Update nodejs-guestbook to be buildable with GCP Buildpacks

briandealwis:
Adds new profile `buildpacks` to build using the [GCP Buildpacks](/GoogleCloudPlatform/buildpacks).

This PR updates the guestbook `package.json` files as their `devDependencies` such as `express` are production-time dependencies.  GCP Buildpacks installs dependencies using `npm ci --quiet` with `NODE_ENV=production`, and so `devDependencies` are not installed.

PR #365, opened 2020-06-26
Make adjustments to Node.js projects for Skaffold buildpack support

j-windsor:
GCP Buildpacks does not install devDependencies. We need to switch to using dependencies instead.


simonz130:
Is this draft ready @j-windsor?

j-windsor:
Looks like #361 made the same changes. Closing now.

Issue #366, opened 2020-06-26
Can't run buildpacks built python containers with `skaffold dev`


matthewmichihara:
`Procfile`s were added to the python samples in order to help with buildpack detection: https://github.com/GoogleCloudPlatform/cloud-code-samples/pull/364

Buildpacks are now detected as builders by `skaffold init --analyze`, but running python sample containers built using buildpacks with `skaffold dev` are running into dependency issues. From the `python-hello-world` sample:

Modified `skaffold.yaml`:
```yaml
apiVersion: skaffold/v2beta5
kind: Config
build:
  tagPolicy:
    sha256: {}
  artifacts:
  - image: python-hello-world
    context: src
    buildpacks:
      builder: gcr.io/buildpacks/builder:v1
deploy:
  kubectl:
    manifests:
    - kubernetes-manifests/**
profiles:
- name: cloudbuild
  build:
    googleCloudBuild: {}
```

```
/Users/michihara/Library/Application Support/cloud-code/bin/versions/8409c589921e86c7502501bcf008deaea7066f73286b757f4c2f083cea56fc28/skaffold dev --filename skaffold.yaml --label ide=Idea --label ideVersion=2020.1.2.0.0 --label ijPluginVersion=unknown --auto-build=false --auto-deploy=false --default-repo gcr.io/chelseamarket --rpc-port 50051 --port-forward=true --status-check=true --enable-rpc=true
Listing files to watch...
 - python-hello-world
Generating tags...
 - python-hello-world -> gcr.io/chelseamarket/python-hello-world:latest
Checking cache...
 - python-hello-world: Found Remotely
Tags used in deployment:
 - python-hello-world -> gcr.io/chelseamarket/python-hello-world:latest@sha256:77a6b5f0ef7ab28aa6ea8e0ce2dc67f88bec5d140c216123f437d6f6013c10a5
Starting deploy...
 - deployment.apps/python-hello-world created
 - service/python-hello-world-external created
Waiting for deployments to stabilize...
 - deployment/python-hello-world: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/python-hello-world-6895bd8b8b-6tzc9: creating container server
 - deployment/python-hello-world: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/python-hello-world-6895bd8b8b-6tzc9: creating container server
 - deployment/python-hello-world: waiting for rollout to finish: 0 of 1 updated replicas are available...
[...]

PR #368, opened 2020-06-29
Add ptvsd to requirements.txt for python samples

matthewmichihara:
Fixes https://github.com/GoogleCloudPlatform/cloud-code-samples/issues/366

This gets the buildpacks builder to install the required ptvsd dependency.


python/cloud-run-django-hello-world/requirements.txt
```diff
@@ -1,2 +1,3 @@
 django==3.0.7
-requests==2.23.0
\ No newline at end of file
+requests==2.23.0
+ptvsd==4.3.2
```
simonz130, line 3: This feels a bit weird to me.
I understand that we need to package this with the code, but this is for debugging purposes rather than one of the app dependencies. Is there another way to bundle it or at the very least, maybe add a comment to explain that this is not part of the "core" app?

matthewmichihara: Sounds good. I don't know another way to bundle `ptvsd` that will be picked up by buildpacks, but I'm also not very experienced with Python.

How does a comment like `This is required for debugging.` sound?

matthewmichihara: Updated with comments

PR #378, opened 2020-07-13
Update host binding for Python Flask app

etanshaul:
From @steren 's friction log:
https://docs.google.com/document/d/18fNxxeRUiYkRmyNJmTIOda2DjzngizJrUJYJOWfdTlk/edit#

The 0.0.0.0 host is printed by flask on app startup, but the url is not reachable in some systems. Updating to use localhost.

### DIFF
--- a/python/cloud-run-python-hello-world/app.py
+++ b/python/cloud-run-python-hello-world/app.py
@@ -38,4 +38,4 @@ def hello():
 
 if __name__ == '__main__':
     server_port = os.environ.get('PORT', '8080')
-    app.run(debug=False, port=server_port, host='0.0.0.0')
+    app.run(debug=False, port=server_port, host='127.0.0.1')


simonz130 (approved):

etanshaul:
@simonz130 any idea why CI is not running?

etanshaul:
@simonz130 looks like CI is failing / not running some jobs, for some unknown reason

simonz130:
I think the CI is failing due to the change in the PR.
This is what the logs say:
```
ERROR: (gcloud.run.deploy) Cloud Run error: Container failed to start. Failed to start and then listen on the port defined by the PORT environment variable. Logs for this revision might contain more information.
```

etanshaul:
looks like this change can't be made. specifically (from https://cloud.google.com/run/docs/troubleshooting):

> A common reason for Cloud Run services failing to start is that the server process inside the container is configured to listen on the localhost (127.0.0.1) address. This refers to the loopback network interface, which is not accessible from outside the container and therefore Cloud Run health check cannot be performed, causing the service deployment failure.

> To solve this, configure your application to start the HTTP server to listen on all network interfaces, commonly denoted as 0.0.0.0.

etanshaul:
closing for now.

Issue #379, opened 2020-07-16
Shorten default project names


kschaab:
In VS Code the default project names are too long, especially for cloud run and take up a lot of space.  This should be shortened as the user has the option of changing the name upon creation.

kschaab:
@russwolf @chairy - The samples for k8s are much smaller, we can remove the cloud-run portion for the cloud run samples, wdyt?

russwolf:
so what are the proposed names? 
k8s-hello-world
k8s-guestbook
cr-hello-world
?

Or something else?

kschaab:
Right now all the k8s names are like nodejs-hello-world.  We can do the same for the Cloud Run samples, so they would be nodejs-hello-world and would collide.  Not what you proposed, @russwolf feel free to lead with a proposal that you think works.

chairy:
I like Russell’s proposal. I don’t think we need the language in the project names.

kschaab:
Updated all the samples to remove language and tech, so `cloud-run-nodejs-hello-world` is now defaulted to `hello-world`.

PR #384, opened 2020-07-19
Shortening default project names per #379

kschaab:
This change is to address #379.  Additionally purging the obsolete template file.


kschaab:
@chairy PTAL.

.templates/template_location_v2.json
```diff
@@ -2,143 +2,143 @@
   {
     "repoPath": "https://github.com/GoogleCloudPlatform/cloud-code-samples.git",
     "directoryPath": "python/python-hello-world",
-    "templatePath": "python-hello-world",
+    "templatePath": "hello-world",
```
simonz130, line 5: This is not going to address the "issue" with the short/long names. This field is a path within the repo to the code. It's not visible to the users but is used for the extension to determine where code of the corresponding sample sits. If you want to change it across the samples, you also need to rename folders, update yamls for deployment and adjust the CI definitions. 

I personally don't think it worth it because we show all the samples for all the languages in the drop down where developers chooses to create a new app. So they might use the drop down twice or more for different languages and then get confused on which sample was cloned. However if the team decides this is the right thing to do for the product, I'll review it and merge when it's ready.

simonz130 (changes requested):

.templates/template_location_v2.json
kschaab: @simonz130 - I'm not sure I understand your feedback.  VS Code uses the folder on disk as the "project name" so shortening the default folder name will thus shorten the name of the project.  The directory path is the path into the repository is it not?

chairy: Did you mean this, @simonz130 ?
image

simonz130: @chairy, yes - this across all of the languages and samples.
@kschaab - are you saying that "templatePath" field is used only to determine the suggested folder where developers will have the template cloned? If so, what will happen if developers will try templates in two different languages for the same deployment platform? Are extensions checking on use-case of existing folder to offer a different folder (like hello-world1, hello-world2, etc.)? Is it the case in both IntelliJ and VSCode?

simonz130: @etanshaul - could you take a look from IntelliJ perspective too?

kschaab: @simonz130 if a user specifies two samples that are the same name they will be presented with a numeric suffix to differentiate them by default.  This only affects the default path that we present to the user, the user is free to type in any path that they would like to save to when we save the sample.

simonz130 (approved):

.templates/template_location_v2.json
simonz130: OK, in that case I'll merge it - at least it will make the default saving directory proposal shorter.

PR #400, opened 2020-08-06
fix project retrieval in node and python messages in Cloud Run samples

j-windsor:
- Set messages to be consistient in python cloud run.
- Get project correctly in node cloud run.


j-windsor:
@simonz130 can I get a merge? I still don't have permission it seems.

PR #409, opened 2020-08-12
Update node to a version supported by the VS Node Debugger

j-windsor:


j-windsor:
Awaiting VSCode v1.49 (August release) to push this change. There is a bug fix there to ensure node 12 properly debugs. Node 8 still does not properly debug in the latest insiders release.

simonz130:
> Awaiting VSCode v1.49 (August release) to push this change. There is a bug fix there to ensure node 12 properly debugs. Node 8 still does not properly debug in the latest insiders release.

Set your PR as draft PR or prefix with [DO NOT MERGE] so we won't merge it before VS releases August bugfix.

j-windsor:
Hey @simonz130. They actually released a patch for this bug, so this PR is ready to merge.

nodejs/nodejs-guestbook/src/backend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use base node 8 image from Docker hub
-FROM node:8
+FROM node:12-alpine
```
kelsk, line 2: Is there a particular reason why we should use `node:12-alpine`? Most of the other GCP Node.js samples are using `node:12-slim`, so it might be preferable to use the same image for consistency.

j-windsor: Good call!

j-windsor:
Thanks for the review @kelsk! PTAL and merge when ready.

PR #416, opened 2020-08-17
Simplify Golang images

briandealwis:
As Cloud Code for VS Code now leverages `skaffold debug` under the hood, this PR re-proposes #102 with some slight updates.

- Add `GOTRACEBACK` env var so skaffold debug can identify images as golang (as per #259)
- Use multi-stage builds to produce smaller images using [distroless/base](https://github.com/GoogleContainerTools/distroless#examples-with-docker)
- Take advantage of Skaffold's `SKAFFOLD_GO_GCFLAGS` to use mode-specific compiler flags, like to disable optimization when using `skaffold debug`.
    - NB: there is a [bug in Skaffold 1.14](https://github.com/GoogleContainerTools/skaffold/issues/4863), [fixed for 1.16 (forthcoming)](https://github.com/GoogleContainerTools/skaffold/pull/4864), that causes the `SKAFFOLD_GO_GCFLAGS` to be ignored by the Go compiler.  This bug does not break Skaffold 1.14 (current version distributed by Cloud SDK).


golang/go-cloud-run-hello-world/Dockerfile
```diff
@@ -16,18 +13,18 @@ COPY . ./
...
-# If you want to run WITHOUT the debugging server:
-#   * COMMENT OUT the previous ENTRYPOINT statements,
-#   * UNCOMMENT the following ENTRYPOINT statement.
-# ENTRYPOINT ["/app"]
+WORKDIR /app
+COPY --from=build /app /app/server
+COPY index.html /app/index.html
+COPY assets /app/assets/
+
+ENTRYPOINT ["/app/server"]
```
simonz130, line 30: Why do we change the entrypoint from /app to /app/server?

briandealwis: It felt wrong to be copying `index.html` and `assets` into the root. It also seemed weird to have `/app/app`.

briandealwis: I'm more than happy to change `/app/server` to something else.  `/deploy/app`?

briandealwis: I changed this to `/hello-world` and kept the binary as `app`.

kelsk: LGTM - @simonz130 any additional feedback?

simonz130 (approved):

PR #419, opened 2020-08-18
Remove redundant dockerfile entrypoint

quoctruong:
Remove the debug entrypoint as we are now using skaffold debug.

### DIFF
--- a/python/python-hello-world/src/Dockerfile
+++ b/python/python-hello-world/src/Dockerfile
@@ -14,4 +14,4 @@ RUN pip install --trusted-host pypi.python.org -r requirements.txt
 COPY . .
 
 # Run app.py when the container launches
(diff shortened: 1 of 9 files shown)


python/python-hello-world/src/Dockerfile
```diff
@@ -14,4 +14,4 @@ RUN pip install --trusted-host pypi.python.org -r requirements.txt
 COPY . .
 
 # Run app.py when the container launches
-ENTRYPOINT ["python", "-m", "ptvsd", "--port", "3000", "--host", "0.0.0.0", "app.py"]
```
matthewmichihara, line 17: Oh nice these aren't required anymore? I think we can also clean up the `Procfile`s as well

matthewmichihara:
@quoctruong can you update the `Procfile`s as well?

quoctruong:
@matthewmichihara Sure. I'll do that in another PR

PR #463, opened 2020-10-06
Update K8s "Hello World" - python

kelsk:


python/python-hello-world/src/app.py
```diff
@@ -1,22 +1,36 @@
...
-    message = "Hello World"
-    return message
+    message = "It's running!"
 
-if __name__ == '__main__':
-    server_port = os.environ.get('PORT')
-    if server_port is None:
-        print("error: PORT environment variable not set")
-        exit(1)
+    return render_template('index.html', message=message)
```
dinagraves, line 32: You're passing in a message to your template, but your template doesn't seem to have any variables to interact with this message?  

Do we need the template?

python/python-hello-world/src/app.py
```diff
@@ -1,22 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 A sample Hello World server.
 """
 import os
-from flask import Flask
+import requests
```
dinagraves, line 19: You don't seem to be using requests?   A Python linter would catch unused imports.

simonz130 (approved):

PR #476, opened 2020-10-16
Remove all trailing commas from .vscode/*.json files

kelsk:


kelsk:
@simonz130 I ran the checks twice: 7 triggers timed out the first time, 12 triggers timed out the second. Could this be an issue with running every trigger simultaneously?

simonz130 (approved):

PR #480, opened 2020-10-16
Skip tests in skaffold dev loop when building with Jib

matthewmichihara:
This matches the behavior when using the Docker builder: https://github.com/GoogleCloudPlatform/cloud-code-samples/blob/master/java/java-hello-world/Dockerfile#L18


kelsk (approved):
LGTM

PR #492, opened 2020-10-27
fix PORT variable usage in test script

### LINKED ISSUE #491: [BUG] python test script uses PORT var incorrectly
**Describe the bug**
The test script `test_content.sh` does not parse the PORT variable correctly, which results in an invalid `$url`. 

see: https://github.com/GoogleCloudPlatform/cloud-code-samples/blob/master/python/cloud-run-python-hello-world/test_content.sh#L18

**To Reproduce**
1. add `echo $url` anywhere in test_content.sh
2. execute the test script, e.g. `bash test_content.sh`

**Expected behavior**
Except output `http://localhost:8080` or similar. 
Actual output: `http://localhost:$PORT`

** Template in use:**
 - Python Flask

**Cloud Code in use:**
 - Cloud Code in Cloud Shell

tjwebb:
resolves #491

### DIFF
--- a/python/cloud-run-python-hello-world/test_content.sh
+++ b/python/cloud-run-python-hello-world/test_content.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 PORT=${PORT:-8080}
-url=${1:-'http://localhost:$PORT'}
+url=${1:-'http://localhost:'$PORT}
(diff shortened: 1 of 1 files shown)


tjwebb:
anybody home?

kelsk (approved):
Thanks for catching this!

PR #495, opened 2020-10-29
Ensure golang images have consistent WORKDIR

briandealwis:
Our golang examples now use a multi-stage build to save approx 800MB of unnecessary baggage (839MB -> 26MB).  But these images currently compile the binary from a different location than where it is run in the result image.  This causes problems for Cloud Code as Golang binaries usually embed the full path of the source.

These patches amend the golang Dockerfiles to use the same WORKDIR for both the build and deployment stages of the images.


kelsk (approved):
LGTM

Issue #415, opened 2020-08-17
[BUG] i/o timeout errors with go-guestbook/backend


briandealwis:
**Describe the bug**
Backend <--> mongodb connection issue results in timeout errors:
```console
$ curl localhost:4503     # frontend
querying backend failed: Get "http://go-guestbook-backend:8080/messages": dial tcp 10.98.193.86:8080: i/o timeout
```

If I query the backend separately:
```
$ curl localhost:8080/messages
curl: (52) Empty reply from server
```

On the main log
```
$ skaffold dev --port-forward
[...]
Waiting for deployments to stabilize...
 - deployment/go-guestbook-frontend: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/go-guestbook-frontend-7b964c645d-bsgrf: creating container frontend
 - deployment/go-guestbook-mongodb: waiting for rollout to finish: 0 of 1 updated replicas are available...
    - pod/go-guestbook-mongodb-7657cd7dcf-srwp8: creating container mongo
 - deployment/go-guestbook-frontend is ready. [2/3 deployment(s) still pending]
 - deployment/go-guestbook-backend is ready. [1/3 deployment(s) still pending]
 - deployment/go-guestbook-mongodb is ready.
Deployments stabilized in 3.914989202s
Port forwarding service/go-guestbook-backend in namespace default, remote port 8080 -> address 127.0.0.1 port 8080
Port forwarding service/go-guestbook-frontend in namespace default, remote port 80 -> address 127.0.0.1 port 4503
Port forwarding service/go-guestbook-mongodb in namespace default, remote port 27017 -> address 127.0.0.1 port 27017
Press Ctrl+C to exit
Watching for changes...
[frontend] API server listening at: [::]:3000
[frontend] 2020/08/17 21:28:17 frontend server listening on port 8080
[backend] API server listening at: [::]:3000
[init-db-ready] Waiting for mongodb at go-guestbook-mongodb:27017 to go live before the BE...
[frontend] 2020/08/17 21:28:22 received request: GET /
[frontend] 2020/08/17 21:28:22 querying backend for entries
[ ... long pause ... ]
[backend] 2020/08/17 21:28:41 backend server listening on port 8080
```

**To Reproduce**
1. Go to 'golang/go-guestbook'
2. Run 'skaffold dev --port-forward
3. Try fetching the forwarded ports for `frontend` or `backend`.

**Expected behavior**

[...]

PR #525, opened 2020-12-07
Add readiness probe for go-guestbook

kelsk:


kelsk:
Fixes #415

donmccasland (approved):
Looks like a readiness probe to me.  is 5 seconds enough for the service to come up?

Issue #385, opened 2020-07-21
[BUG] Flask-based python-* apps do not terminate gracefully


briandealwis:
**Describe the bug**
When using `skaffold dev` with the python-hello-world app, it takes 30 seconds before a new redeployed container starts contributing output.  The underlying cause is that the [Flask dev server ignores `SIGTERM`](https://github.com/docker/compose/issues/4199) and instead only responds to `SIGINT`, the signal sent by Kubernetes to gracefully terminate a container.

**To Reproduce**
1. Go to `python/python-hello-world`
2. Run `skaffold dev --port-forward`
3. Run `while true; do date; curl http://localhost:4503; sleep 5; done` (or whatever port you receive)
4. Change the source.
5. See that it takes 30s at least before the change takes effect.

**Expected behavior**
Like the other languages, changes should be reflected within 3-4 seconds — which is also too long, but that's a different issue!

**Possible Fixes**

1. **Set a `terminationGracePeriodSeconds` with a small number.**  This is a bandaid and does not actually cleanly shutdown the server.  It is not a practice we should advocate.
2. **Send `kill SIGINT` from a container `preStop` lifecycle-hook.**  This cleanly terminates the server. But `preStop` feels like it is meant for a band-aid for pulling in existing applications rather than new greenfield development.
3. **Use gunicorn.** The Flask dev server is not meant for production usage and we shouldn't encourage its use in our examples, especially since they are used as suggested templates.
4. **Add a SIGINT handler.**

I think the third option is the right option.  But the fourth is easiest.

PR #527, opened 2020-12-07
Adding SIGINT handler

donmccasland:
Addressing #385 by adding SIGINT handler


dinagraves:
Gunicorn isn't supported by `skaffold debug` so we're going to use this solution

kelsk (approved):
✅ tested deploy & debug on vs code

PR #684, opened 2021-05-06
Configure go-guestbook root module

kelsk:
The current structure of go-guestbook defines the `backend` and `frontend` services as go modules. This causes errors when using the [gopls](https://github.com/golang/tools/blob/master/gopls/doc/workspace.md) IDE hover feature, which requires one go module in the root folder of the open workspace.

To fix this, I have:
- moved go.mod & go.sum to the root folder
- updated the skaffold.yaml to utilize the beta `docker` flag, and set the context as the root folder
- updated the frontend & backend Dockerfiles to access the `go.*` files in the root folder


sivakku:
Thanks @kelsk. Change LGTM!

sivakku (approved):
LGTM

PR #715, opened 2021-05-19
Add timeouts to cloudbuild config files

kelsk:
Adding the following timeouts to remove test failures due to default 10 minute timeouts:

- build - 2500s
- step: 'Deploy to staging' - 1200s
- step: 'Get Endpoint' - 1200s


dinagraves (approved):
LGTM! 

However, 20 minute timeouts for a single step seems a little long, although it doesn't seem to be ever taking up the whole time. 

At some point, we should look into optimizing these builds to speed them up.

kelsk:
> At some point, we should look into optimizing these builds to speed them up.

Agreed! The 20 minute timeout covers builds that occasionally run > 15 minutes. Refactoring the builds is on the todo list.

PR #731, opened 2021-06-02
Fix Java image and service names

Walliee:


kelsk (approved):

kelsk:
@Walliee Thanks for catching this!

PR #854, opened 2021-10-20
Update Skaffold schema version

kelsk:
- Updates the schema version of all Skaffold files to ~`v2beta22`~ `v2beta19`
- Changes `buildpack` to `buildpacks`

Since the `master` branch is deprecated, the schema version is set to the version compatible with Cloud Code effective August 2021.

### DIFF
--- a/dotnet/dotnet-guestbook/skaffold.yaml
+++ b/dotnet/dotnet-guestbook/skaffold.yaml
@@ -1,6 +1,6 @@
 # To learn more about the skaffold.yaml schema visit
 # https://skaffold.dev/docs/references/yaml/
-apiVersion: skaffold/v2beta16
+apiVersion: skaffold/v2beta19
 kind: Config
 build:
   tagPolicy:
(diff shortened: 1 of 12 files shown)


etanshaul:
Thanks for fixing this! One consideration is that IntelliJ users are slower to upgrade plugin versions, and we pin to a specific Skaffold binary. Is August when IntelliJ release a version of CC compatible with this updated schema version? I think supporting the current release and the previous two is reasonable.

kelsk:
> Thanks for fixing this! One consideration is that IntelliJ users are slower to upgrade plugin versions, and we pin to a specific Skaffold binary. Is August when IntelliJ release a version of CC compatible with this updated schema version? I think supporting the current release and the previous two is reasonable.

Ah, good call. I'll revert a few more versions back to `v2beta16`, which is compatible for VS Code back to June. I assume it's compatible to a similar time for IntelliJ, though I don't know how to check for certain with IntelliJ. (VS Code has a hardcoded list of their schema versions.)

kelsk:
> I think `v2beta19` is a good compromise of ~2 months of IDE support having been out. wdyt?

That sounds good to me! I'll update the `v1` branch to be `v2beta19` as well (since it's currently at `18`).

And thanks for showing me how to find the skaffold version and use it to identify the schema version. I'll use it to find the corresponding schema for future samples version releases.

briandealwis (approved):

etanshaul (approved):

PR #985, opened 2022-03-14
Update image links to point to correct location

davidponder:
Currently, when Cloud Run samples are checked out via Cloud Code for VS Code, they open a preview of `.readmes/vscode/README.md`. Unfortunately, the images in this file don't load, and this leads to users seeing broken image links. This PR updates the image file paths so that viewing them in VS Code works.

Before:
image

After: 
image

I tried to follow the same logic seen on the Kubernetes versions of these READMEs.


kelsk (approved):
@davidponder thank you for catching and fixing this!

PR #1103, opened 2022-08-12
Node.js: use Alpine base image

ace-n:


averikitsch (approved):

iennae:
just curious how this speeds up CI for node going from slim to alpine. isn't slim the smallest form factor? or are we spending time installing git and bash so this ultimately speeds up CI? (https://hub.docker.com/_/node) It's not obvious from the PR so figure you have more insight.

ace-n:
As discussed in our team chatroom - `alpine` images are smaller than `slim` ones, so they _should_ be faster to unpack.

Furthermore - I confirmed that an example `alpine` build outran its `slim` counterpart in [this PR](https://github.com/GoogleCloudPlatform/cloud-code-samples/pull/1096).

PR #1130, opened 2022-09-19
Fix cloudcode 'run individual services' to include having a running backend, where needed

sushicw:
Some of the samples require a running backend, this adds a step to make sure the user runs once with cleanup=false so that a running backend can be present while updating just the frontend. 

Checked the other guestbook samples and verified they are robust to no backend (and thus don't need the extra step). 

Fixes internal bug b/233103747


iennae (approved):

sushicw:
@iennae Can this please be merged? Thanks!

PR #1143, opened 2022-10-11
Add stub for GCF hello world sample

glouischandra:


iennae (approved):
LGTM; needs testing added, but we can mere this for now.

PR #1148, opened 2022-10-27
Fix Java sample

davidponder:
Adds a pom and properly nests the file based on our other samples.

Heavily based on creating a new Cloud Function via the Cloud Console UI.


frostie (approved):
LGTM

PR #1224, opened 2023-03-16
Add dockerfile profile to java launch.config files to fix k8s deployments

SKrupa:
Currently, Cloud Code defaults to using cloudbuild if a user opens samples.workspace and tries to deploy the java samples to k8s due to maven changes (see: https://github.com/GoogleCloudPlatform/cloud-code-samples/pull/1207)

### DIFF
--- a/java/java-hello-world/.vscode/launch.json
+++ b/java/java-hello-world/.vscode/launch.json
@@ -9,9 +9,11 @@
         "type": "cloudcode.kubernetes",
         "request": "launch",
         "skaffoldConfig": "${workspaceFolder}/skaffold.yaml",
+        "profile": "dockerfile",
         "watch": true,
         "cleanUp": true,
-        "portForward": true
+        "portForward": true,
+        "imageRegistry": "gcr.io/cloud-code-vs-code"
       }
     ]
   }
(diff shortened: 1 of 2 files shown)


davidponder (approved):

java/java-hello-world/.vscode/launch.json
```diff
@@ -9,9 +9,11 @@
         "type": "cloudcode.kubernetes",
         "request": "launch",
         "skaffoldConfig": "${workspaceFolder}/skaffold.yaml",
+        "profile": "dockerfile",
         "watch": true,
         "cleanUp": true,
-        "portForward": true
+        "portForward": true,
+        "imageRegistry": "gcr.io/cloud-code-vs-code"
```
davidponder: I don't think this should be filled in (users can specify their own imageRegistry i think)

davidponder (changes requested):

davidponder (approved):

averikitsch (approved):
Ultimately, I would like to use the cloudbuild profile, because Jib is a better containerization method than a Dockerfile. However, that is dependent on a maintained Cloud Builders with both Skaffold and Maven installed.

PR #1230, opened 2023-04-06
style: Rename django-python-hello-world samples to be unique

SKrupa:
Currently they use the same name as the existing python-hello-world sample. As these samples are deploying different code, they should use different names. This is consistent with how the guestbook samples are defined, where the django sample use a different name than the base python ones.


python/django/python-hello-world/kubernetes-manifests/hello.deployment.yaml
```diff
@@ -6,25 +6,25 @@
 apiVersion: apps/v1
```
averikitsch, line 6: Missing update on line 2

averikitsch (changes requested):

python/django/python-hello-world/kubernetes-manifests/hello.deployment.yaml
SKrupa: whoops, fixed. Thanks!

averikitsch (approved):

PR #1248, opened 2023-06-15
Set CGO_ENABLED=0 for distroless builds

j-windsor:
https://github.com/GoogleContainerTools/distroless/issues/1342 breaks go samples due to a mismatch of c library availability between distroless images and goland images. Setting CGO_ENABLED=0 bypasses the need for these libs.


davidponder (approved):

iennae (approved):
LGTM

PR #1249, opened 2023-06-16
Set CGO_ENABLED=0 for the whole build to be set on all go commands

j-windsor:
After #1248, we are still seeing some breakages on the `go mod download`. This sets the variable for the whole build so any call to `go` will get the option.

### DIFF
--- a/golang/go-cloud-run-hello-world/Dockerfile
+++ b/golang/go-cloud-run-hello-world/Dockerfile
@@ -2,6 +2,9 @@
 FROM golang:1.19 AS build
 
 WORKDIR /hello-world
+
+# Avoid dynamic linking of libc, since we are using a different deployment image
+# that might have a different version of libc.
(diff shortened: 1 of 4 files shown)


briandealwis (changes requested):
Could you put a comment before each occurrence to link to the issue explaining the problem?

Or alternatively could just say something like `avoid dynamic linking of libc`. 

And two other changes:
- since we’re not linking against libc then we can use the `static` images
- distroless now publishes distribution tags (`static-debian11`)

See example [here](https://github.com/GoogleContainerTools/distroless#examples-with-docker).

iennae (approved):
LGTM

ref doc for image change - https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md 

static is the right one for use with Go.

briandealwis (approved):
LGTM

